PR #3, opened 2020-10-26
add scc and remove configmap

zhiweiyin318:
1. add scc for OCP (use vendor label of managedcluster to get the type of mangedcluster)
2. remove configmap and configure submariner automatically in OCP

related task: https://github.com/open-cluster-management/backlog/issues/6432


pkg/helpers/helpers.go
```diff
@@ -188,3 +189,28 @@ func GetBrokerTokenAndCA(client kubernetes.Interface, brokerNS, clusterName stri
...
+var (
+	clusterInfoGVR schema.GroupVersionResource = schema.GroupVersionResource{
+		Group:    "internal.open-cluster-management.io",
+		Version:  "v1beta1",
+		Resource: "managedclusterinfos",
+	}
+)
+
+func GetClusterType(dynamicClient dynamic.Interface, clusterName string) (string, error) {
+	clusterInfo, err := dynamicClient.Resource(clusterInfoGVR).Namespace(clusterName).Get(context.TODO(), clusterName, metav1.GetOptions{})
```
qiujian16, line 202: you can get that from vendor label in managedCluster. do you?

zhiweiyin318: yes, I can get vendor label in managedcluster, but the label may be customized by users.

skeeey: I think we should reduce the dependence on other apis, do we need consider if there is no ACM installed? can we provide a flag (or an env) for submariner-addon, we may try to get the cluster type by `managedclusterinfos` api firstly, if it fails, we try to get the cluster type from the flag? or keep this simple, we get the cluster type from vendor label and doc this

qiujian16: i would go with label only. We might have some api changes on managedcluster later  to include more informations, but label is good enough is the  first step.

zhiweiyin318: done

PR #19, opened 2020-11-12
support to install by olm

zhiweiyin318:


Makefile
```diff
@@ -82,6 +82,12 @@ verify-scripts:
 .PHONY: verify-scripts
 verify: verify-scripts verify-codegen-crds
 
+deploy-operator: ensure-operator-sdk
```
skeeey, line 85: could we change this task name? e.g. deploy-with-olm or else, this name may not be straightforward

zhiweiyin318: done

PR #27, opened 2020-12-12
Add apiVersion string to the ManagedClusterSet resource in README

giannisalinetti:
Add apiVersion string in the README section named "Setup Submariner on the Hub cluster", step 1. This avoid validation errors when submitting the resource in the hub cluster.


qiujian16:
/approve
/lgtm

Thanks!

PR #29, opened 2020-12-17
Using olm catalog to deploy submariner

skeeey:


skeeey:
we need wait for Submariner v0.8.0 release

skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/7536

deploy/config/crds/submariner.io_clusters_crd.yaml
```diff
@@ -1,13 +1,63 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1alpha1
```
qiujian16, line 1: is this v1alpha1 or v1?

pkg/hub/submarineragent/controller.go
```diff
@@ -302,3 +336,76 @@ func (c *submarinerAgentController) removeClusterRBACFiles(ctx context.Context,
...
+	clusterType, err := helpers.GetClusterType(clusterClient, config.ClusterName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the cluster type %+v : %+v", config.ClusterName, err)
+	}
+	switch clusterType {
+	case helpers.ClusterTypeOCP:
+		//TODO: the scc should be handled in submariner v0.8.0, but need verify
+		config.NATEnabled = true
+		if version.IsDownStream() {
+			config.CatalogSource = "redhat-operators"
```
qiujian16, line 405: I  think we only  need to set different catalog source when it is in downstream. We may not set an explicit env to say it is downstream

pkg/hub/submarineragent/controller.go
```diff
@@ -302,3 +336,76 @@ func (c *submarinerAgentController) removeClusterRBACFiles(ctx context.Context,
...
+		}
+		config.CatalogSourceNamespace = "openshift-marketplace"
```
qiujian16, line 407: i think this should also be set differently in downstream?

skeeey: it needn't, all catalogsources are in this namespace on ocp

pkg/hub/submarineragent/controller.go
skeeey: good suggestion

deploy/config/crds/submariner.io_clusters_crd.yaml
zhiweiyin318: yes, the CRD has been updated to apiextensions.k8s.io/v1.
suggest to get the CRD from submariner-operator repo by using `make` command.  cc @skeeey

skeeey: ok, I will

zhiweiyin318:
@skeeey  `deploy/olm-catalog` and `csv` should be updated by `make update-csv` too.

skeeey:
> @skeeey `deploy/olm-catalog` and `csv` should be updated by `make update-csv` too.

done

deploy/config/crds/submariner.io_clusters_crd.yaml
skeeey: done

PR #34, opened 2021-01-11
remove submariner CRDs from csv bundle

zhiweiyin318:
relates issue: https://github.com/open-cluster-management/backlog/issues/8367


pkg/helpers/helpers.go
```diff
@@ -125,6 +130,65 @@ func UpdateSubmarinerConfigStatusFn(cond metav1.Condition, managedClusterInfo co
 	}
 }
 
+type CRDClientHolder struct {
+	apiExtensionsClient apiextensionsclient.Interface
+}
+
+func NewCRDClientHolder() *CRDClientHolder {
```
qiujian16, line 137: why we just use clientholder in library go directly

pkg/hub/submarinerbroker/brokerCRDsController.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	return operatorhelpers.NewMultiLineAggregate(errs)
+}
+
+func (c *submarinerBrokerCRDsController) cleanUp(ctx context.Context) error {
+	errs := []error{}
+	for _, crd := range brokerCRDNames {
+		err := c.crdClient.ApiextensionsV1().CustomResourceDefinitions().Delete(ctx, crd, metav1.DeleteOptions{})
```
qiujian16, line 118: do we get into this code? the crd is remove only when operator is deleted from olm.  Will olm guarantee the sequence that crd is delete at first, then operator?

pkg/helpers/helpers.go
```diff
@@ -125,6 +130,65 @@ func UpdateSubmarinerConfigStatusFn(cond metav1.Condition, managedClusterInfo co
...
+	return &CRDClientHolder{}
+}
+func (c *CRDClientHolder) WithAPIExtensionsClient(client apiextensionsclient.Interface) *CRDClientHolder {
+	c.apiExtensionsClient = client
+	return c
+}
+
+// ApplyCRDDirectly is used to apply CRD v1 and v1beta1 since resourceapply in library-go cannot apply CRD v1 with error
+func ApplyCRDDirectly(
```
qiujian16, line 146: we can  live  with this, but we should submit a pr  to library-go.

pkg/hub/submarinerbroker/brokerCRDsController.go
skeeey: not sure...

pkg/helpers/helpers.go
zhiweiyin318: `apiExtensionsClient` in `clientholder` is an internal, cannot call outside.

pkg/hub/submarinerbroker/brokerCRDsController.go
zhiweiyin318: I think CRD is the last one to be deleted in uninstall Operator process (refer https://github.com/operator-framework/olm-book/blob/master/docs/uninstall-an-operator.md)  
cleanup is not needed here.

PR #53, opened 2021-04-26
Using managedclusteraddon to deploy submariner-addon

skeeey:


skeeey:
see https://github.com/open-cluster-management/backlog/issues/6440

pkg/hub/manager.go
```diff
@@ -79,22 +90,31 @@ func RunControllerManager(ctx context.Context, controllerContext *controllercmd.
...
+	go addOnInformers.Start(ctx.Done())
 	go submarinerBrokerCRDsController.Run(ctx, 1)
 	go submarinerBrokerController.Run(ctx, 1)
 	go submarinerAgentController.Run(ctx, 1)
 
+	mgr, err := addonmanager.New(controllerContext.KubeConfig)
+	if err != nil {
+		return err
+	}
+	mgr.AddAgent(submarineraddonagent.NewAddOnAgent(rand.String(5), kubeClient, controllerContext.EventRecorder))
```
qiujian16, line 115: using rand here might be a problem. It will change after each restart

pkg/hub/submarineraddonagent/agent.go
```diff
@@ -0,0 +1,189 @@
...
+func NewAddOnAgent(agentName string, kubeClient kubernetes.Interface, recorder events.Recorder) *addOnAgent {
+	return &addOnAgent{
+		agentName:  agentName,
+		kubeClient: kubeClient,
+		recorder:   recorder,
+	}
+}
+
+// Manifests generates manifestworks to deploy the submariner-addon agent on the managed cluster
+func (a *addOnAgent) Manifests(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn) ([]runtime.Object, error) {
```
qiujian16, line 75: I think you would also need to create submariner sa here?

pkg/hub/submarineragent/controller.go
```diff
@@ -231,22 +252,27 @@ func (c *submarinerAgentController) syncManagedCluster(
 	ctx context.Context,
 	managedCluster *clusterv1.ManagedCluster,
 	config *configv1alpha1.SubmarinerConfig) error {
-	// the cluster does not have the submariner label, try to clean up the submariner agent
-	if _, existed := managedCluster.Labels[submarinerLabel]; !existed {
-		return c.cleanUpSubmarinerAgent(ctx, managedCluster)
-	}
-
-	// the cluster does not have the clusterset label, try to clean up the submariner agent
 	clusterSetName, existed := managedCluster.Labels[clusterSetLabel]
```
qiujian16, line 255: do we generete clusterset claim in managedcluster? we should rely on that instead of calling clusterset on hub.

pkg/hub/submarineraddonagent/agent.go
```diff
@@ -0,0 +1,189 @@
...
+	clusterAddOnGroup := fmt.Sprintf("system:open-cluster-management:cluster:%s:addon:%s", cluster.Name, addOnName)
+	if !requestingOrgs.Has(clusterAddOnGroup) {
+		return false
+	}
+
+	if !strings.HasPrefix(x509cr.Subject.CommonName, clusterAddOnGroup+":agent:") {
+		return false
+	}
+
+	return csr.Spec.Username == x509cr.Subject.CommonName
```
qiujian16, line 159: in current implementation, i do not think this will be equal. @elgnay is it correct?

elgnay: No. the user name of csr is different with common name in cert. In the current implementation, csr for addon registration is always created with hub-kubeconfig of registration-agent.

pkg/hub/manager.go
skeeey: instead of using a fixed agent name

pkg/hub/submarineraddonagent/agent.go
skeeey: we don't need a submariner sa here, it will be created by hub controller

pkg/hub/submarineragent/controller.go
skeeey: This is hub part, so we don't need to generate claim

pkg/hub/submarineraddonagent/agent.go
skeeey: yes, you are right, I check the csr request common name here

qiujian16 (commented):
/approve
/lgtm

PR #55, opened 2021-04-28
Make sure that the addon agent image is same with hub image and change the addon name to `submariner`

skeeey:

### DIFF
--- a/pkg/hub/submarineraddonagent/agent.go
+++ b/pkg/hub/submarineraddonagent/agent.go
@@ -36,17 +36,20 @@ func init() {
 }
 
 const (
-	addOnName                    = "submariner-addon"
+	addOnName                    = "submariner"
 	agentName                    = "submariner-addon-agent"
+	defaultAgentImage            = "quay.io/open-cluster-management/submariner-addon:latest"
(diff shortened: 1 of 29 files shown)


pkg/hub/submarineraddonagent/agent.go
```diff
@@ -187,3 +208,18 @@ func (a *addOnAgent) permissionConfig(cluster *clusterv1.ManagedCluster, addon *
 
 	return operatorhelpers.NewMultiLineAggregate(errs)
 }
+
+func (a *addOnAgent) getCurrentPodImage() (string, error) {
+	podNamespce := os.Getenv("POD_NAMESPACE")
```
qiujian16, line 213: i think we only needs to call this once before starting controller. And i think it is helpful if we have an agent-image flag. If it  is not set, we fall back to use this method to get image. if it fails,  we fatal the whole container.

skeeey: done

skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/6440

qiujian16 (commented):
/lgtm

PR #62, opened 2021-05-11
fix the condition is wrong when there are no connections

skeeey:


qiujian16 (commented):
/lgtm

PR #65, opened 2021-05-14
update to submairner 0.9.0

skeeey:


test/integration/crds/submariner/submariner.io_submariners.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
```
zhiweiyin318, line 4: need to change the dir name 0.8.1 to 0.9.0 ?

PR #85, opened 2021-07-15
Use go:embed instead of go-bindata

skitt:
This drops a direct dependency and removes a lot of boilerplate code.


qiujian16:
/approve
/hold

hold until we switch to 2.4, thanks!

qiujian16:
@skitt would you like to rebase this and i think it is good to go

/unhold

PR #98, opened 2021-08-11
Convert the GCP unit tests to use the Gingko framework

tpantelis:
See individual commits for details.


skeeey:
@tpantelis thanks for your contribution, but for gcp part, we will use https://github.com/submariner-io/cloud-prepare to replace the current implementation, so I think you may add this part to the cloud-prepare lib

PR #106, opened 2021-08-17
Use the Submariner API modules

skitt:
Instead of depending on all of Submariner and its operator, pull in
their new API modules only; this simplifies the dependency tree (which
is required to pull in the cloud-prepare library without creating a
mess of conflicting dependencies).


tpantelis:
It seems this PR also removed the vendor dir - was this intended (it wasn't mentioned in the PR summary)?

tpantelis (approved):

PR #110, opened 2021-08-24
Enable PR dependency check jobs

Jaanki:
This job marks a PR `dependent` whenever a dependent PR
is mentioned via `Depends on/ depends on` keyword.

The job fails until the dependent PR is not merged. This
helps in properly managing merging of dependent PRs.

Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

### DIFF
--- a/.github/workflows/dependent-issues.yml
+++ b/.github/workflows/dependent-issues.yml
@@ -1,0 +1,38 @@
+---
+name: PR Dependencies
+
+on:
+  issues:
+    types:
+      - opened
+      - edited
+      - closed
+      - reopened
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - closed
+      - reopened
+      - synchronize
+  schedule:
+    - cron: '0 0/6 * * *'  # every 6 hours
+
+jobs:
+  check:
+    name: Check Dependencies
+    if: github.repository_owner == 'open-cluster-management'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: z0al/dependent-issues@65a8dadac8eb87a2ea821f277fe5529d86609558
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # The label to use to mark dependent issues
+          label: dependent
+
+          # Enable checking for dependencies in issues.
+          check_issues: on
+
+          # A comma-separated list of keywords to mark dependency.
+          keywords: depends on, Depends on


Jaanki:
Depends on #1

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,39 @@
...
+  schedule:
+    - cron: '0 0/6 * * *'  # every 6 hours
+
+jobs:
+  check:
+    name: Check Dependencies
+    if: github.repository_owner == 'submariner-io'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: z0al/dependent-issues@1208b4dd978973b14de5b0b227a2871dbbf34ffb
```
dfarrell07, line 28: We should either use the same SHA (1.3.0 release) as the [other repos](https://github.com/submariner-io/submariner-operator/blob/devel/.github/workflows/dependent-issues.yml#L28) or use [just-out 1.4.1](https://github.com/z0al/dependent-issues/releases) SHA.

dfarrell07 (changes requested):

.github/workflows/dependent-issues.yml
Jaanki: @dfarrell07 I don't see the difference between the SHAs (1.3.0) as is in other repos.

skitt: @dfarrell07 I confirm the SHA is correct (and isn’t 1.3.0’s); but it would be better to use 65a8dadac8eb87a2ea821f277fe5529d86609558 (the 1.4.1 SHA).

skitt:
/approve cancel

(Line 25 needs to be fixed.)

.github/workflows/dependent-issues.yml
dfarrell07: Ah because we're using an old version before they introduced the bug related to Dependabot. Might be good to mention in the commit message, but otherwise :+1:

dfarrell07: I see 1.4.1 contains your PR to fix their Dependabot bug @skitt, that's awesome! :partying_face: Do I understand https://github.com/z0al/dependent-issues/pull/227/commits/ae5a6e10286604c302839c58072d591115ffb440 correctly that the default will now make things work as desired for us? If so, +1 to using that SHA.

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,38 @@
...
+  schedule:
+    - cron: '0 0/6 * * *'  # every 6 hours
+
+jobs:
+  check:
+    name: Check Dependencies
+    if: github.repository_owner == 'open-cluster-management'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: z0al/dependent-issues@65a8dadac8eb87a2ea821f277fe5529d86609558
```
Jaanki, line 27: Using v1.4.1 SHA

.github/workflows/dependent-issues.yml
Jaanki: @skitt 's change matters if using `pull+request` event. We are using `pull_request_target` event. So it shouldn't affect us. Anyways, using 1.4.1 SHA just because its the latest.

skitt: @Janki not quite, my change matters in both cases. It does affect us, by restoring the old behaviour — so dependabot PRs still get processed by dependent-issues (which wasn’t the case with 1.3.0).

Jaanki: ahh...from the README, I thought its only for `pull_request`

PR #112, opened 2021-08-24
Use cloud-prepare for AWS provisioning

skitt:
This is a first step towards delegating all cloud provisioning to
cloud-prepare. For now, we only set up the cloud-prepare client; all
the actual AWS configuration still relies on the addon-specific code.

The Submariner dependencies need to be updated to the current devel
versions (with the associated transitive dependencies) to work with
the K8s 1.19+ APIs.


pkg/cloud/aws/aws.go
```diff
@@ -63,9 +67,11 @@ type awsProvider struct {
 	clusterName   string
 	instanceType  string
 	gateways      int
+	cloudPrepare  cpapi.Cloud
 }
 
 func NewAWSProvider(
+	restConfig *rest.Config,
```
tpantelis, line 74: Why is this passed in? It looks unused. `rest.Config` is problematic for unit tests in terms of mocking so would be ideal if we can avoid passing it in code we want to unit test.

tpantelis (changes requested):

pkg/cloud/aws/aws.go
skitt: Ah yes, thanks, it was needed with a previous version of cloud-prepare (with the MachineSetDeployer) but is no longer necessary.

tpantelis (approved):

PR #113, opened 2021-08-24
Update the connections status controller unit tests

tpantelis:
Modified to use Gingko and also added more test cases and more extensive verifications. Also fixed issues found or other refactoring. See individual commits for details.


dfarrell07:
Testing if the ci-operator is already accessible:

 /test golangci-lint

PR #117, opened 2021-08-30
Add NATTDiscoveryPort to submarinerconfig spec

vthapar:

### DIFF
--- a/deploy/config/crds/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
+++ b/deploy/config/crds/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
@@ -40,6 +40,10 @@ spec:
                 default: 4500
                 description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
                 type: integer
+              NATTDiscoveryPort:
+                default: 4900
+                description: NATTDiscoveryPort represents NAT-T Discovery port (default UDP/4900).
+                type: integer
               NATTEnable:
                 default: true
                 description: NATTEnable represents IPsec NAT-T enabled (default true).
--- a/deploy/olm-catalog/manifests/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
+++ b/deploy/olm-catalog/manifests/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
@@ -38,6 +38,10 @@ spec:
                 default: 4500
                 description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
                 type: integer
+              NATTDiscoveryPort:
+                default: 4900
+                description: NATTDiscoveryPort represents NAT-T Discovery port (default UDP/4900).
+                type: integer
               NATTEnable:
                 default: true
                 description: NATTEnable represents IPsec NAT-T enabled (default true).
--- a/pkg/apis/submarinerconfig/v1alpha1/0000_00_submarineraddon.open-cluster-management.io_submarinerconfigs.crd.yaml
+++ b/pkg/apis/submarinerconfig/v1alpha1/0000_00_submarineraddon.open-cluster-management.io_submarinerconfigs.crd.yaml
@@ -40,6 +40,10 @@ spec:
                 default: 4500
                 description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
                 type: integer
+              NATTDiscoveryPort:
+                default: 4900
+                description: NATTDiscoveryPort represents NAT-T Discovery port (default UDP/4900).
+                type: integer
               NATTEnable:
                 default: true
                 description: NATTEnable represents IPsec NAT-T enabled (default true).
--- a/pkg/cloud/aws/aws.go
+++ b/pkg/cloud/aws/aws.go
@@ -56,17 +56,18 @@ type machineSetConfig struct {
 }
 
 type awsProvider struct {
-	workClinet    workclient.Interface
-	awsClinet     client.Interface
-	eventRecorder events.Recorder
-	region        string
-	infraId       string
-	ikePort       int64
-	nattPort      int64
-	clusterName   string
-	instanceType  string
-	gateways      int
-	cloudPrepare  cpapi.Cloud
+	workClinet        workclient.Interface
+	awsClinet         client.Interface
+	eventRecorder     events.Recorder
+	region            string
+	infraId           string
+	ikePort           int64
+	nattPort          int64
+	nattDiscoveryPort int64
+	clusterName       string
+	instanceType      string
+	gateways          int
+	cloudPrepare      cpapi.Cloud
 }
 
 func NewAWSProvider(
@@ -74,7 +75,7 @@ func NewAWSProvider(
 	eventRecorder events.Recorder,
 	region, infraId, clusterName, credentialsSecretName string,
 	instanceType string,
-	ikePort, nattPort int,
+	ikePort, nattPort, nattDiscoveryPort int,
 	gateways int) (*awsProvider, error) {
 	if region == "" {
 		return nil, fmt.Errorf("cluster region is empty")
@@ -96,6 +97,10 @@ func NewAWSProvider(
 		nattPort = helpers.SubmarinerNatTPort
 	}
 
+	if nattDiscoveryPort == 0 {
+		nattDiscoveryPort = helpers.SubmarinerNatTDiscoveryPort
+	}
+
 	if instanceType == "" {
 		instanceType = defaultInstanceType
 	}
@@ -106,17 +111,18 @@ func NewAWSProvider(
 	}
 
 	return &awsProvider{
-		workClinet:    workClient,
-		awsClinet:     awsClient,
-		eventRecorder: eventRecorder,
-		region:        region,
-		infraId:       infraId,
-		ikePort:       int64(ikePort),
-		nattPort:      int64(nattPort),
-		clusterName:   clusterName,
-		instanceType:  instanceType,
-		gateways:      gateways,
-		cloudPrepare:  cpaws.NewCloud(awsClient.FullEC2API(), infraId, region),
+		workClinet:        workClient,
+		awsClinet:         awsClient,
+		eventRecorder:     eventRecorder,
+		region:            region,
+		infraId:           infraId,
+		ikePort:           int64(ikePort),
+		nattPort:          int64(nattPort),
+		nattDiscoveryPort: int64(nattDiscoveryPort),
+		clusterName:       clusterName,
+		instanceType:      instanceType,
+		gateways:          gateways,
+		cloudPrepare:      cpaws.NewCloud(awsClient.FullEC2API(), infraId, region),
 	}, nil
 }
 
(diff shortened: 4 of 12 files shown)


deploy/olm-catalog/0.3.0/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
```diff
@@ -35,6 +35,9 @@ spec:
             IPSecNATTPort:
               description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
               type: integer
+            NATTDiscoveryPort:
+              description: NATTDiscoveryPort represents NAT-T Discovery port (default 4900).
```
sridhargaddam, line 39: same

deploy/olm-catalog/manifests/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
```diff
@@ -38,6 +38,10 @@ spec:
                 default: 4500
                 description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
                 type: integer
+              NATTDiscoveryPort:
+                default: 4900
+                description: NATTDiscoveryPort represents NAT-T Discovery port (default 4900).
```
sridhargaddam, line 43: same

pkg/apis/submarinerconfig/v1alpha1/0000_00_submarineraddon.open-cluster-management.io_submarinerconfigs.crd.yaml
```diff
@@ -40,6 +40,10 @@ spec:
                 default: 4500
                 description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
                 type: integer
+              NATTDiscoveryPort:
+                default: 4900
+                description: NATTDiscoveryPort represents NAT-T Discovery port (default 4900).
```
sridhargaddam, line 45: same

pkg/cloud/aws/aws.go
```diff
@@ -56,25 +56,26 @@ type machineSetConfig struct {
...
-	clusterName   string
-	instanceType  string
-	gateways      int
-	cloudPrepare  cpapi.Cloud
+	workClinet        workclient.Interface
+	awsClinet         client.Interface
+	eventRecorder     events.Recorder
+	region            string
+	infraId           string
+	ikePort           int64
```
sridhargaddam, line 64: ikePort can be removed as we are not longer using it. But it can be done in a separate PR.

pkg/cloud/gcp/gcp.go
```diff
@@ -72,11 +78,13 @@ func NewGCPProvider(
...
 //    - IPsec IKE port (by default 500/UDP)
 //    - NAT traversal port (by default 4500/UDP)
+//    - NAT traversal discovery port (by default 4900/UDP)
 //    - 4800/UDP port to encapsulate Pod traffic from worker and master nodes to the Submariner Gateway nodes
 // 2. create the inbound and outbound firewall rules to open 8080/TCP port to export metrics service from the Submariner gateway
 func (g *gcpProvider) PrepareSubmarinerClusterEnv() error {
-	// open IPsec IKE port (by default 500/UDP), NAT traversal port (by default 4500/UDP) and route port (4800/UDP)
-	ports := []string{g.ikePort, g.nattPort, g.routePort}
+	// open IPsec IKE port (by default 500/UDP), NAT traversal port (by default 4500/UDP),
+	// NAT Traversal discovery port (be default 4900/udp) and route port (4800/UDP)
```
sridhargaddam, line 86:
```suggestion
	// NAT Traversal discovery port (by default 4900/udp) and vx-submariner tunnel port (4800/UDP)
```

scripts/deploy.sh
```diff
@@ -211,7 +211,8 @@ EOF
 {"op":"add","path":"/spec/imagePullSpecs/submarinerImagePullSpec","value":"'${submrepo}'/submariner-gateway:'${submver}'"},
 {"op":"add","path":"/spec/imagePullSpecs/submarinerRouteAgentImagePullSpec","value":"'${submrepo}'/submariner-route-agent:'${submver}'"},
 {"op":"add","path":"/spec/imagePullSpecs/lighthouseAgentImagePullSpec","value":"'${submrepo}'/lighthouse-agent:'${submver}'"},
-{"op":"add","path":"/spec/imagePullSpecs/lighthouseCoreDNSImagePullSpec","value":"'${submrepo}'/lighthouse-coredns:'${submver}'"}]'
+{"op":"add","path":"/spec/imagePullSpecs/lighthouseCoreDNSImagePullSpec","value":"'${submrepo}'/lighthouse-coredns:'${submver}'"},
+{"op":"add","path":"/spec/NATTEnable","value":false}]'
```
sridhargaddam, line 215: Is this change related to this PR? It looks like its for NAT enable/disable support.

sridhargaddam (commented):
Not completely familiar with the code. But from a high level, it looks good. Some minor nits.

pkg/cloud/gcp/gcp.go
vthapar: It uses the term route port in code too, so I'd prefer to retain it to avoid confusion.

skitt: There’s still the “be” typo ;-).

deploy/olm-catalog/0.3.0/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
```diff
@@ -35,6 +35,9 @@ spec:
             IPSecNATTPort:
               description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
               type: integer
+            NATTDiscoveryPort:
+              description: NATTDiscoveryPort represents NAT-T Discovery port (default UDP/4900).
+              type: integer
```
skitt, line 40: I don’t think the 0.3.0 catalog entry should be changed.

skitt (changes requested):

deploy/config/crds/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
```diff
@@ -40,6 +40,10 @@ spec:
                 default: 4500
                 description: IPSecNATTPort represents IPsec NAT-T port (default 4500).
                 type: integer
+              NATTDiscoveryPort:
+                default: 4900
+                description: NATTDiscoveryPort represents NAT-T Discovery port (default UDP/4900).
```
mangelajo, line 45: In a future this may need to be a list, for example for on-premise deployments, where each gateway (unless it hass a public IP of it's own) needs a separate port.

Do you think it could be doable to start using a list now?

Also IPSEC port has the same issue, so may be it's simpler to do it the way you do it now, and then in the future when we support on-prem better with this add-on you move both fields into lists.

pkg/cloud/aws/aws.go
```diff
@@ -56,25 +56,26 @@ type machineSetConfig struct {
...
-	eventRecorder events.Recorder
-	region        string
-	infraId       string
-	ikePort       int64
-	nattPort      int64
-	clusterName   string
-	instanceType  string
-	gateways      int
-	cloudPrepare  cpapi.Cloud
+	workClinet        workclient.Interface
```
mangelajo, line 59:
```suggestion
	workClient        workclient.Interface
```
small typo

pkg/cloud/aws/aws.go
```diff
@@ -106,25 +111,26 @@ func NewAWSProvider(
...
-		eventRecorder: eventRecorder,
-		region:        region,
-		infraId:       infraId,
-		ikePort:       int64(ikePort),
-		nattPort:      int64(nattPort),
-		clusterName:   clusterName,
-		instanceType:  instanceType,
-		gateways:      gateways,
-		cloudPrepare:  cpaws.NewCloud(awsClient.FullEC2API(), infraId, region),
+		workClinet:        workClient,
```
mangelajo, line 114:
```suggestion
		workClient:        workClient,
```

pkg/cloud/aws/aws.go
```diff
@@ -106,25 +111,26 @@ func NewAWSProvider(
...
+		cloudPrepare:      cpaws.NewCloud(awsClient.FullEC2API(), infraId, region),
 	}, nil
 }
 
 // PrepareSubmarinerClusterEnv prepares submariner cluster environment on AWS
 // The below tasks will be executed
 // 1. open submariner route port (4800/UDP) between all master and worker nodes
 // 2. open submariner metrics port (8080/TCP) between all master and worker nodes
-// 3. open IPsec ports (by default, 4500/UDP and 500/UDP) for submariner gateway instances
+// 3. open IPsec ports (by default, 4500/UDP, 4900/UDP and 500/UDP) for submariner gateway instances
```
mangelajo, line 133:
```suggestion
// 3. open IPsec and NAT discovery ports (by default, 4500/UDP, 4900/UDP and 500/UDP) for submariner gateway instances
```

pkg/cloud/aws/aws.go
```diff
@@ -160,7 +166,7 @@ func (a *awsProvider) PrepareSubmarinerClusterEnv() error {
 		return fmt.Errorf("failed to open route port in security group: %v \n", err)
 	}
 
-	// Open IPsec ports (by default, 4500/UDP and 500/UDP) for submariner gateway instances
+	// Open IPsec ports (by default, 4500/UDP, 4900/UDP and 500/UDP) for submariner gateway instances
```
mangelajo, line 169:
```suggestion
	// Open IPsec and NAT discovery ports (by default, 4500/UDP, 4900/UDP and 500/UDP) for submariner gateway instances
```

mangelajo (changes requested):
some typo, and a small comment about long term that probably doesn't need action right now. May be adding a comment somewhere?

deploy/config/crds/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
vthapar: There is a way we can make it work without needing it to be a list. For multiple gateways we can treat configured port as base port and increment by 1 for every extra gateway. e.g. in config we only have 4500 for IPSEC, but if user configures 3 gateways we can use 4500, 4501 and 4502. Such a change can be easily implemented on submariner code without needing changes anywhere else.

If above doesn't work, we can always make it a user configured list later.

deploy/olm-catalog/0.3.0/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
vthapar: Yeah, I wasn't sure of this either, but this change was already done in fix for VxLan and NattEnable. Should we remove those from 0.3.0 too?

pkg/cloud/aws/aws.go
vthapar: @mangelajo This is used in about 20 different places in code. IMO this change should be done in a separate PR to avoid cluttering this one. thoughts?

deploy/olm-catalog/0.3.0/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
skitt: Ah yes, I’ve pushed https://github.com/open-cluster-management/submariner-addon/pull/126 to address that.

aswinsuryan (approved):

skitt (approved):

PR #120, opened 2021-08-31
Add golangci-lint make target, linter config

dfarrell07:
Add a make target for running Go linting via golangci-lint.

Add configuration for golangci-lint, copied from the main Submariner
repo. Disable all currently-failing tests.

This will be consumed by Prow to run CI. The tool install will be
handled by Prow.

To use locally, golangci-lint and go >1.15 are required.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,90 @@
+---
+linters-settings:
+  gocritic:
+    enabled-tags:
+      - diagnostic
+      - opinionated
+      # - performance
+      - style
+    disabled-checks:
+      - ifElseChain
+      - unnamedResult
+  gocyclo:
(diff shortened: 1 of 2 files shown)


dfarrell07:
When I try to run this locally, it shows this warning (which I don't understand) and passes (which clearly seems like a false positive). Any suggestions?

> WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package : could not load export data: no export data for "embed"

dfarrell07:
This works now, thanks for the Go version suggestion @skitt!

dfarrell07:
(verifying the assign plugin is enabled for this repo)

/assign @skitt

PR #137, opened 2021-09-21
Use the cloud-prepare GCP client interface

aswinsuryan:
Signed-off-by: Aswin Surayanarayanan <asuryana@redhat.com>


pkg/cloud/gcp/gcp.go
```diff
@@ -46,24 +65,50 @@ func NewGCPProvider(
...
+	restMapper, err := buildRestMapper(restConfig)
 	if err != nil {
 		return nil, err
 	}
 
+	msDeployer := ocp.NewK8sMachinesetDeployer(restMapper, dynamicClient)
+
+	k8sClient, _ := k8s.NewK8sInterface(kubeClient)
+
+	gwDeployer, err := cloudpreparegcp.NewOcpGatewayDeployer(cloudPrepare, msDeployer, instanceType,
```
sridhargaddam, line 99: error check missing

pkg/hub/submarineragent/controller.go
```diff
@@ -375,24 +379,33 @@ func (c *submarinerAgentController) syncSubmarinerConfig(ctx context.Context,
 
 	// prepare submariner cluster environment
 	errs := []error{}
-	cloudProvider, preparedErr := cloud.GetCloudProvider(c.restConfig, c.kubeClient, c.manifestWorkClient, c.dynamicClient,
-		c.eventRecorder, managedClusterInfo, config)
-	if preparedErr == nil {
-		preparedErr = cloudProvider.PrepareSubmarinerClusterEnv()
-	}
+	var condition  metav1.Condition
+	if config.Status.ManagedClusterInfo.Platform == "AWS" {
```
sridhargaddam, line 383: This is checking only for AWS. What about GCP/VmWare? Why are we handling AWS and GCP (or VmWare) separately? I feel we they should be handled at the same place.

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -121,29 +121,34 @@ func NewSubmarinerConfigController(
 }
 
 func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	klog.Errorf("Syncing in  %s in %s",  c.clusterName, helpers.SubmarinerAddOnName)
```
sridhargaddam, line 124: Errorf?

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -121,29 +121,34 @@ func NewSubmarinerConfigController(
...
 		// the config not found, could be deleted, do nothing
+		klog.Errorf("Config not found in  %s in %s",  c.clusterName, helpers.SubmarinerConfigName)
 		return nil
 	}
 	if err != nil {
+		klog.Errorf("Error not nil  %s in %s",  c.clusterName, helpers.SubmarinerConfigName)
 		return err
 	}
-
+	klog.Errorf("The managed platform is %s", config.Status.ManagedClusterInfo.Platform)
```
sridhargaddam, line 146: ditto and below

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -368,15 +402,8 @@ func (c *submarinerConfigController) unlabelNode(ctx context.Context, config *co
...
-	// currently only gcp is supported
-	switch config.Status.ManagedClusterInfo.Platform {
-	case "GCP":
-		zoneLabel = gcpZoneLabel
-	default:
-		// for other non-public cloud platform (vsphere) or native k8s
-		zoneLabel = defaultZoneLabel
-	}
+	// for other non-public cloud platform (vsphere) or native k8s
+	zoneLabel := defaultZoneLabel
```
sridhargaddam, line 406: I think we can implement similar functionality in cloud-prepare where we just tag nodes as Gatewaynodes and assume that the firewall is properly configured out of band. Once it is implemented, can we remove this code and use cloud-prepare?

sridhargaddam (changes requested):
@aswinsuryan thanks for the PR. Please split the PR into individual commits (at least the vendor files can be independent commit IMO)

pkg/cloud/cloud.go
```diff
@@ -57,10 +58,14 @@ func GetCloudProvider(
...
 		return gcp.NewGCPProvider(
+			restConfig,
 			kubeClient,
+			dynamicClient,
+			hubKubeClient,
 			eventsRecorder,
-			infraId, clusterName, config.Spec.CredentialsSecret.Name,
-			config.Spec.IPSecNATTPort, config.Spec.NATTDiscoveryPort,
+			region, infraId, clusterName, config.Spec.CredentialsSecret.Name,
+			config.Spec.GatewayConfig.AWS.InstanceType,
```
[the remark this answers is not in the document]
aswinsuryan: Using "" now , we don't have GCP instance type now accepted from the UI, once that is added we can pass that.

pkg/spoke/submarineragent/config_controller.go
aswinsuryan: yes that can be done. This code is retained just for vmware. GCP/AWS does not use it now.

pkg/spoke/submarineragent/config_controller.go
aswinsuryan: Debug code, removed.

pkg/spoke/submarineragent/config_controller.go
aswinsuryan: Removed.

pkg/cloud/cloud.go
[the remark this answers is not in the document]
sridhargaddam: Please note that we need the GatewayInstanceType only when we are interested in creating dedicated gateway nodes, otherwise existing worker nodes will be labelled. So, you can just pass "" for now since you are not really creating dedicated gateway nodes.

pkg/hub/submarineragent/controller.go
```diff
@@ -371,23 +377,37 @@ func (c *submarinerAgentController) syncSubmarinerConfig(ctx context.Context,
...
+	} else {
+		if !helpers.IsSubmarinerEnvPrepared(c.configClient, config.Namespace, config.Name) {
+			condition = metav1.Condition{
+				Type:   configv1alpha1.SubmarinerConfigConditionEnvPrepared,
+				Status: metav1.ConditionFalse,
+				Reason: "SubmarinerClusterEnvPreparationNotYetDone",
+			}
+		} else {
+			return operatorhelpers.NewMultiLineAggregate(errs)
+		}
```
tpantelis, line 410: I don't think we should or need to update the condition here to indicate it's pending.  The spoke is responsible for  updating the status in this case. Also we could get a race where the pending status overrides the completed status.   

```suggestion
		return nil
```

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -181,6 +189,46 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
 		return c.updateGatewayStatus(syncCtx.Recorder(), config)
 	}
 
+	if config.Status.ManagedClusterInfo.Platform == "GCP" {
+		if !helpers.IsSubmarinerEnvPrepared(c.configClient, config.Namespace, config.Name) {
```
tpantelis, line 193: This check doesn't seem necessary.

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -181,6 +189,46 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
...
+
+			if preparedErr != nil {
+				return preparedErr
+			} else {
+				preparedErr = cloudProvider.PrepareSubmarinerClusterEnv()
+			}
+
+			if preparedErr != nil {
+				return preparedErr
+			}
```
tpantelis, line 205: We should be consistent with the `submarinerAgentController` which reports the error in the status condition.

tpantelis (changes requested):

pkg/cloud/gcp/gcp.go
```diff
@@ -1,43 +1,62 @@
...
+	metricsPort       uint16
+	cloudPrepare      api.Cloud
 	eventRecorder     events.Recorder
+	gwDeployer        api.GatewayDeployer
+	gateways          int
+	nattDiscoveryPort int64
 }
 
 func NewGCPProvider(
+	restConfig *rest.Config,
```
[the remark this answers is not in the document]
aswinsuryan: Changed to RESTMapper

pkg/cloud/gcp/gcp.go
```diff
@@ -100,101 +155,73 @@ func (g *gcpProvider) PrepareSubmarinerClusterEnv() error {
...
+	if err != nil {
+		return err
 	}
-
 	return operatorhelpers.NewMultiLineAggregate(errs)
 }
 
-func (g *gcpProvider) openPorts(ingress, egress *compute.Firewall) error {
-	var errs []error
+func NewClient(kubeClient kubernetes.Interface, secretNamespace, secretName string) (string, gcpclient.Interface, error) {
```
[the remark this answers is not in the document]
aswinsuryan: done

pkg/hub/submarineragent/controller.go
```diff
@@ -371,23 +377,37 @@ func (c *submarinerAgentController) syncSubmarinerConfig(ctx context.Context,
 
 	// prepare submariner cluster environment
 	errs := []error{}
-	cloudProvider, preparedErr := cloud.GetCloudProvider(c.kubeClient, c.manifestWorkClient, c.eventRecorder, managedClusterInfo, config)
-	if preparedErr == nil {
-		preparedErr = cloudProvider.PrepareSubmarinerClusterEnv()
-	}
+	var condition metav1.Condition
+	if config.Status.ManagedClusterInfo.Platform != "GCP" {
+		cloudProvider, preparedErr := cloud.GetCloudProvider(c.restConfig, c.kubeClient, c.manifestWorkClient, c.dynamicClient,
```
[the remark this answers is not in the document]
aswinsuryan: fixed, now using nil value.

pkg/hub/submarineragent/controller.go
aswinsuryan: Now using nil value for conditon.

pkg/spoke/submarineragent/config_controller.go
aswinsuryan: Done.

pkg/hub/submarineragent/controller.go
```diff
@@ -568,16 +588,20 @@ func (c *submarinerAgentController) cleanUpSubmarinerClusterEnv(ctx context.Cont
...
 		//TODO handle the error gracefully in the future
 		c.eventRecorder.Warningf("CleanUpSubmarinerClusterEnvFailed", "failed to create cloud provider: %v", err)
 		return nil
 	}
-	if err := cloudProvider.CleanUpSubmarinerClusterEnv(); err != nil {
-		//TODO handle the error gracefully in the future
-		c.eventRecorder.Warningf("CleanUpSubmarinerClusterEnvFailed", "failed to clean up cloud environment: %v", err)
-		return nil
+
+	if config.Status.ManagedClusterInfo.Platform != "GCP" {
```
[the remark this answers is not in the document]
aswinsuryan: Removed.

pkg/hub/submarineragent/controller.go
```diff
@@ -371,23 +377,37 @@ func (c *submarinerAgentController) syncSubmarinerConfig(ctx context.Context,
 
 	// prepare submariner cluster environment
 	errs := []error{}
-	cloudProvider, preparedErr := cloud.GetCloudProvider(c.kubeClient, c.manifestWorkClient, c.eventRecorder, managedClusterInfo, config)
-	if preparedErr == nil {
-		preparedErr = cloudProvider.PrepareSubmarinerClusterEnv()
-	}
+	var condition metav1.Condition
+	if config.Status.ManagedClusterInfo.Platform != "GCP" {
```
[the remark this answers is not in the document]
aswinsuryan: we need do any for GCP , but just update the status with managedClusterInfo which is required in spoke.

pkg/spoke/submarineragent/config_controller.go
tpantelis: This was marked as resolved but wasn't. Please remove this call and the function.

pkg/hub/submarineragent/controller.go
```diff
@@ -371,23 +375,27 @@ func (c *submarinerAgentController) syncSubmarinerConfig(ctx context.Context,
 
 	// prepare submariner cluster environment
 	errs := []error{}
-	cloudProvider, preparedErr := cloud.GetCloudProvider(c.kubeClient, c.manifestWorkClient, c.eventRecorder, managedClusterInfo, config)
-	if preparedErr == nil {
-		preparedErr = cloudProvider.PrepareSubmarinerClusterEnv()
-	}
+	var condition *metav1.Condition
+	if managedClusterInfo.Platform != "GCP" {
```
tpantelis, line 379: Instead of falling thru as a no-op and modifying `UpdateSubmarinerConfigStatusFn` to handle nil condition, just return here:

```suggestion
	if managedClusterInfo.Platform == "GCP" {
	    return nil
	}
```

tpantelis (changes requested):

pkg/hub/submarineragent/controller.go
aswinsuryan: it is not a no-op , helpers.UpdateSubmarinerConfigStatusFn(condition, managedClusterInfo), the managedclusterInfo is updated and is required in spoke

tpantelis: oh right.

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -154,6 +152,16 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
...
 		}
 
+		if config.Status.ManagedClusterInfo.Platform == "GCP" {
+			cloudProvider, preparedErr := cloud.GetCloudProvider(c.restMapper, c.kubeClient, nil, c.dynamicClient,
+				c.hubKubeClient, c.eventRecorder, config.Status.ManagedClusterInfo, config)
+			if preparedErr == nil {
+				preparedErr = cloudProvider.CleanUpSubmarinerClusterEnv()
+			}
+
+			return nil
```
tpantelis, line 162: Returning here will skip removing the gateways below. I think we only want to return on error.

```suggestion
			if preparedErr != nil {
			    return preparedErr
			{
```

pkg/helpers/helpers.go
```diff
@@ -119,17 +119,36 @@ func UpdateSubmarinerConfigStatus(
 	return updatedStatus, updated, err
 }
 
+func IsSubmarinerEnvPrepared(
```
tpantelis, line 122: This is now only called by the spoke controller.  I'd prefer to move it local rather than lumping it in a general helpers "grab bag" (I'm gradually moving functions out of here).

That said, is it really necessary in the spoke controller?

tpantelis (changes requested):

pkg/spoke/submarineragent/config_controller.go
aswinsuryan: done.

pkg/helpers/helpers.go
aswinsuryan: changed to local. Ya check is required to prevent cloud-prepare being called for each reconcile.

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	k8s.io/client-go v0.21.3
 	k8s.io/code-generator v0.21.3
 	k8s.io/component-base v0.21.3
+	k8s.io/klog v1.0.0
```
skitt, line 35: The project already imports `klog/v2`, so we should avoid importing `klog`.

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -181,6 +191,49 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
 		return c.updateGatewayStatus(syncCtx.Recorder(), config)
 	}
 
+	if config.Status.ManagedClusterInfo.Platform == "GCP" {
+		if IsSubmarinerEnvPrepared(c.configClient, config.Namespace, config.Name) {
```
tpantelis, line 195: `IsSubmarinerEnvPrepared` can be simplified to:

```suggestion
		if meta.IsStatusConditionTrue(config.Status.Conditions, configv1alpha1.SubmarinerConfigConditionEnvPrepared) {
```

tpantelis (changes requested):

pkg/cloud/gcp/gcp.go
```diff
@@ -46,24 +62,50 @@ func NewGCPProvider(
...
+	if err != nil {
+		klog.Errorf("Unable to retrieve the gcpclient :%v", err)
+		return nil, err
+	}
+
+	cloudPrepare := cloudpreparegcp.NewCloud(projectId, infraId, region, gcpClient)
+
+	if err != nil {
+		return nil, err
+	}
```
skitt, line 87: This isn’t needed, `err` hasn’t changed since the last check.

pkg/cloud/gcp/gcp.go
```diff
@@ -100,101 +151,59 @@ func (g *gcpProvider) PrepareSubmarinerClusterEnv() error {
...
-			Direction: "EGRESS",
-			Allowed: []*compute.FirewallAllowed{
-				{
-					IPProtocol: protocol,
-					Ports:      ports,
-				},
-			},
-		}
+func (g *gcpProvider) Started(message string, args ...interface{}) {
+	klog.Infof(fmt.Sprintf(message, args...))
```
skitt, line 196: This should use the event recorder instead (see the AWS PR for an example). The event recorder feeds the logs and the event journal in OCP.

skitt (changes requested):

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -181,6 +191,49 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
 		return c.updateGatewayStatus(syncCtx.Recorder(), config)
 	}
 
+	if config.Status.ManagedClusterInfo.Platform == "GCP" {
+		if meta.IsStatusConditionTrue(config.Status.Conditions, configv1alpha1.SubmarinerConfigConditionEnvPrepared) {
```
tpantelis, line 195: Sorry my original suggestion was incorrect.

```suggestion
		if meta.IsStatusConditionFalse(config.Status.Conditions, configv1alpha1.SubmarinerConfigConditionEnvPrepared) {
```

tpantelis (changes requested):

pkg/cloud/gcp/gcp.go
aswinsuryan: Done

go.mod
aswinsuryan: Done.

pkg/spoke/submarineragent/config_controller.go
aswinsuryan: Done.

pkg/hub/submarineragent/controller.go
```diff
@@ -588,21 +576,20 @@ func (c *submarinerAgentController) cleanUpSubmarinerClusterEnv(ctx context.Cont
 	}
 
 	managedClusterInfo := config.Status.ManagedClusterInfo
-	cloudProvider, err := cloud.GetCloudProvider(c.restConfig, c.kubeClient, c.manifestWorkClient, c.dynamicClient,
+	cloudProvider, err := cloud.GetCloudProvider(nil, c.kubeClient, c.manifestWorkClient, c.dynamicClient,
 		nil, c.eventRecorder, managedClusterInfo, config)
 	if err != nil {
 		//TODO handle the error gracefully in the future
 		c.eventRecorder.Warningf("CleanUpSubmarinerClusterEnvFailed", "failed to create cloud provider: %v", err)
```
sridhargaddam, line 583: Incase you happen to respin the PR. 

"failed to get cloud provider"

sridhargaddam (approved):
Thanks for addressing the comments @aswinsuryan

pkg/hub/submarineragent/controller.go
aswinsuryan: done

tpantelis:
lgtm. @skitt please re-review.

skitt (approved):

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -181,6 +191,49 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
...
+
+			if updatedErr != nil {
+				return err
+			}
+			if updated {
+				c.eventRecorder.Eventf("SubmarinerClusterEnvPrepared", "submariner cluster environment was prepared for manged cluster %s", config.Namespace)
+			}
+		}
+
+		updatedErr := c.updateGatewayStatus(syncCtx.Recorder(), config)
```
tpantelis, line 233: @aswinsuryan @sridhargaddam @skitt Looking at this code in more detail, I notice that we call `updateGatewayStatus` on L191 above for AWS and return.  For GCP, we now also call `updateGatewayStatus` and return here so now we never get to `ensureGateways` below to actually label gateway nodes even for GCP. This doesn't seem right or am I missing something?

aswinsuryan: @tpantelis  The gateway nodes are now labelled by cloud-prepare for AWS and GCP. The `ensureGateways` retained for other clouds like VMware and IBM cloud for tagging the node. In VMware/IBM cloud `SubmarinerAddon` does not program firewall rules and expect the user to do it.

tpantelis: So it seems that if we have a cloud provider implementation for a platform, we can assume it's the cloud provider's responsibility to label the gateways?

 Another thing is the the hub controller checks if `config.Spec.CredentialsSecret` is nil otherwise it skips the cloud preparation.  It seems this should as well. I've added that check locally in the `cloud.ProviderFactory` added by https://github.com/open-cluster-management/submariner-addon/pull/149. However I wonder if we should the check in the spoke controller prior to code freeze.

aswinsuryan: The managedClusterInfo will be created only config.Spec.CredentialsSecret is not nil in the hub. Spoke will respond only if managedClusterInfo is available. SO it should safe to assume config.Spec.CredentialsSecret will be set for GCP.

aswinsuryan: > So it seems that if we have a cloud provider implementation for a platform, we can assume it's the cloud provider's responsibility to label the gateways?

Yes, cloud prepare should do that when we call ocpGatewayDeployer.Deploy()
> 
> Another thing is that the hub controller checks if `config.Spec.CredentialsSecret` is nil otherwise it skips the cloud preparation. It seems spoke controller should as well or is it safe to assume `config.Spec.CredentialsSecret` will be set for GCP? I've added that check in the `cloud.ProviderFactory` added by #149 however I'm sure if we should add the check in the spoke controller prior to code freeze to be safe.

tpantelis: Right, it does AWS and GCP. Ideally this code shouldn't have knowledge of the internal implementation of specific cloud providers. We should just make gateway deployment a requirement and move the gateway labeling done here into a default cloud provider.

tpantelis: > The managedClusterInfo will be created only config.Spec.CredentialsSecret is not nil in the hub. Spoke will respond only if managedClusterInfo is available. SO it should safe to assume config.Spec.CredentialsSecret will be set for GCP.

Actually the hub controller updates the `managedClusterInfo` regardless, ie even if `config.Spec.CredentialsSecret` is nil.

```
            if config.Spec.CredentialsSecret == nil {
		// no platform credentials, the cluster env is not requred to prepare, only update the manged cluster info
		_, _, err := helpers.UpdateSubmarinerConfigStatus(
			c.configClient,
			config.Namespace, config.Name,
			helpers.UpdateSubmarinerConfigStatusFn(&metav1.Condition{...}, managedClusterInfo),
		)
		return err
	}
```

But #149 is merged now.

PR #140, opened 2021-09-21
Update the submarinerBrokerController and unit tests

tpantelis:
Modified the unit tests to use Gingko and also added more test cases and more extensive verifications. Also did some refactoring, simplification and minor fixes. See individual commits for details.

### DIFF
--- a/pkg/finalizer/finalizer.go
+++ b/pkg/finalizer/finalizer.go
@@ -1,0 +1,62 @@
+package finalizer
+
+import (
+	"context"
+
+	"github.com/submariner-io/admiral/pkg/resource"
+	"github.com/submariner-io/admiral/pkg/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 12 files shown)


tpantelis:
@skitt Some jobs are failing verifying dependencies:

```
LC_COLLATE=C diff --no-dereference -N "/tmp/tmp.V9rRzeqbDT"/{current,updated}/go.mod || ( echo '`go.mod` content is incorrect - did you run `go mod tidy`?' && false )
18a19
> 	github.com/submariner-io/admiral v0.11.0-m2
`go.mod` content is incorrect - did you run `go mod tidy`?
```

The admiral dependency was added previously - I'm just using it in this PR.  I pushed [this commit](https://github.com/open-cluster-management/submariner-addon/pull/140/commits/78e935d9d5c6721c92b005751ad15f8a46485f76) that fixes it - not exactly sure why though.

Jaanki (approved):

PR #141, opened 2021-09-22
Add a ManifestWork-based MachineSet deployer

skitt:
This is required to manage deployments through the hub.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/prometheus/procfs v0.7.2 // indirect
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
+	github.com/submariner-io/admiral v0.11.0-m2
 	github.com/submariner-io/cloud-prepare v0.11.0-m2.0.20210920144917-0b33718646a7
 	github.com/submariner-io/submariner-operator/apis v0.0.0-20210817145008-861856b068a1
 	github.com/submariner-io/submariner/pkg/apis v0.0.0-20210817085048-59f656555db0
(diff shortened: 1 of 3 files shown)


pkg/cloud/msd/workmftmsd.go
```diff
@@ -0,0 +1,107 @@
+package work_mft_msd
```
tpantelis, line 1: This seems a bit cryptic plus I don't think the linters like `_` in a package name.  How about we name this 

```suggestion
package machineset
```

and ....

pkg/cloud/msd/workmftmsd.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+type workManifestMachineSetDeployer struct {
+	client        workclient.Interface
+	workName      string
+	clusterName   string
+	eventRecorder events.Recorder
+}
+
+func NewWorkManifestMachineSetDeployer(client workclient.Interface, workName, clusterName string, eventRecorder events.Recorder) ocp.MachineSetDeployer {
```
tpantelis, line 28: ... and name this:

```suggestion
func NewDeployer(client workclient.Interface, workName, clusterName string, eventRecorder events.Recorder) ocp.MachineSetDeployer {
```

tpantelis (changes requested):

pkg/cloud/msd/workmftmsd.go
tpantelis: Or `NewWorkManifestDeployer` if that makes it clearer.

tpantelis (approved):

PR #142, opened 2021-09-22
Delegate cloud preparation on AWS to cloud-prepare

skitt:
This correctly implements cloud preparation on AWS with cloud-prepare,
using a WorkManifest-based MachineSet deployer.


pkg/cloud/cloud.go
```diff
@@ -23,8 +25,10 @@ type CloudProvider interface {
 }
 
 func GetCloudProvider(
+	restConfig *rest.Config,
 	kubeClient kubernetes.Interface,
 	workClient workclient.Interface,
+	dynamicClient dynamic.Interface,
```
tpantelis, line 31: Why were these 2 params added? They don't appear to be used.

pkg/cloud/aws/machineset/machineset.go
```diff
@@ -0,0 +1,17 @@
+package machineset
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+//go:generate mockgen -source=./machineset.go -destination=./mock/machineset_generated.go -package=mock
+
+// Interface wraps an actual machineset deployer for mocking
+type Interface interface {
```
tpantelis, line 8: This looks identical to the [MachineSetDeployer](https://github.com/open-cluster-management/submariner-addon/blob/8a75e04271d64acd77d3d001b2baafac464ccc81/vendor/github.com/submariner-io/cloud-prepare/pkg/ocp/machinesets.go#L36:6) interface in cloud-prepare. What is this used for?

skitt: It’s only used for `mockgen`. Once the tests are moved (or re-implemented) in `cloud-prepare`, all this will go away.

pkg/cloud/cloud.go
skitt: Left-overs from previous iterations where they were used, thanks for spotting them!

pkg/cloud/aws/machineset/machineset.go
tpantelis: I see.  Looking at the usage in _aws_test.go_, two of the test cases only assert that `Deploy` is called with any arg. This is rather minimal and I don't think it's a compelling enough reason to temporarily redefine the interface and generate a mock. I think we should just create a no-op implementation in _aws_test.go_ for now:

```
type fakeMachinesetDeployer struct {
}

func (f fakeMachinesetDeployer) Deploy(machineSet *unstructured.Unstructured) error {
	return nil
}

func (f fakeMachinesetDeployer) GetWorkerNodeImage(machineSet *unstructured.Unstructured, infraID string) (string, error) {
	return "", nil
}

func (f fakeMachinesetDeployer) Delete(machineSet *unstructured.Unstructured) error {
	return nil
}
```

tpantelis (changes requested):

pkg/cloud/aws/machineset/machineset.go
skitt: Good idea, I’ve just done that.

aswinsuryan (approved):

tpantelis (approved):

PR #146, opened 2021-09-23
Move helpers.ApplyManifestWork to manifestwork.Apply

tpantelis:
It's better to avoid lumping unrelated functions into util or helpers packages so move this function to its own package.

Also modified it to use admiral's `CreateOrUpdate`.

### DIFF
--- a/pkg/cloud/manifestwork/machineset.go
+++ b/pkg/cloud/manifestwork/machineset.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"context"
 
-	"github.com/open-cluster-management/submariner-addon/pkg/helpers"
+	"github.com/open-cluster-management/submariner-addon/pkg/manifestwork"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/cloud-prepare/pkg/ocp"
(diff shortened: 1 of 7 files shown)


Jaanki (approved):

aswinsuryan (approved):

sridhargaddam (approved):

skitt (changes requested):
Since this has a conflict which needs to be resolved, may I suggest moving `machineset.go` to the new package location while we’re at it?

tpantelis:
> Since this has a conflict which needs to be resolved, may I suggest moving `machineset.go` to the new package location while we’re at it?

yeah I debated co-locating the two. I could've put `Apply` in _.../cloud/manifestwork_ but it's more generalized and not related to "cloud". OTOH, it seems `NewMachineSetDeployer` is specific to cloud? If not, I'm fine with moving it.

skitt:
> yeah I debated co-locating the two. I could've put `Apply` in _.../cloud/manifestwork_ but it's more generalized and not related to "cloud". OTOH, it seems `NewMachineSetDeployer` is specific to cloud? If not, I'm fine with moving it.

Well, I guess the whole `MachineSetDeployer` idea is `cloud`-specific; but what bothered me really is having the same package names in two different import paths.

tpantelis:
> > yeah I debated co-locating the two. I could've put `Apply` in _.../cloud/manifestwork_ but it's more generalized and not related to "cloud". OTOH, it seems `NewMachineSetDeployer` is specific to cloud? If not, I'm fine with moving it.
> 
> Well, I guess the whole `MachineSetDeployer` idea is `cloud`-specific; but what bothered me really is having the same package names in two different import paths.

I think that's fine - it's not uncommon. There are many util/helper packages out there under different paths.

skitt (approved):

PR #147, opened 2021-09-23
Move to v1 for CRDs

Jaanki:
...instead of v1beta1 which is deprecated from
k8s v1.16+ and unavailable in v1.22+

### DIFF
--- a/pkg/hub/submarinerbroker/manifests/submariner.io_clusters_crd.yaml
+++ b/pkg/hub/submarinerbroker/manifests/submariner.io_clusters_crd.yaml
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   name: clusters.submariner.io
   ownerReferences:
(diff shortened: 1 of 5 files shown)


deploy/olm-catalog/0.3.0/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
skitt, line 1: The 0.3.0 catalog is historical, please don’t change it.

skitt (changes requested):
Thanks for this, it will remove a bunch of warnings when running the addon!

Just one change: please don’t edit the 0.3.0 catalog.

PR #153, opened 2021-09-27
Upgarde open-cluster-management-io/api/client/cluster/clientset

Jaanki:
...to use v1beta1 as v1aplha1 is deprecated as per

https://github.com/open-cluster-management-io/registration/commit/6f8ca332f9d8871b3f8e56917cd5babdfe9b314c


skitt (changes requested):
Please fix the commit message: “use v1beta1 as v1alpha1” instead of “use v11 as v1aplha1”.

skitt:
This should also update references to v1alpha1 ManagedClusterSet APIs in the code, _e.g._ in the hub `manager.go`.

vendor/modules.txt
```diff
@@ -1214,13 +1214,17 @@ open-cluster-management.io/api/client/cluster/clientset/versioned/typed/cluster/
 open-cluster-management.io/api/client/cluster/clientset/versioned/typed/cluster/v1/fake
 open-cluster-management.io/api/client/cluster/clientset/versioned/typed/cluster/v1alpha1
```
Jaanki, line 1215: @skitt  Should I remove these v1alpha1 references manually?

skitt: No, the whole `vendor/modules.txt` is automatically managed by `go mod vendor`; if the entries are still present, something (a dependency) is still using them.

skitt: In fact it’s the OCM API itself which uses them, since it still supports the v1alpha1 API.

skitt (approved):

PR #160, opened 2021-09-30
Only install the APIs we really use ...

skitt:
... instead of installing all the K8s APIs referenced by the OpenShift
API helpers.


skitt:
/hold

This needs to be tested in a real deployment.

tpantelis (approved):

dfarrell07 (approved):

PR #178, opened 2021-10-07
Enable golangci-lint goconst linter, fix errors

dfarrell07:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -45,7 +45,7 @@ linters:
     # - gochecknoglobal # TODO: Do we want this?
     # - gochecknoinits # TODO: Do we want this?
     # - gocognit # Actually runs revive linter
-    # - goconst # TODO: Fix and enable
+    - goconst
     - gocritic
     # - gocyclo # TODO: Do we want this?
     # - godot # TODO: Do we want this?
--- a/pkg/spoke/submarineragent/config_controller_test.go
+++ b/pkg/spoke/submarineragent/config_controller_test.go
@@ -29,6 +29,8 @@ import (
 )
 
 const (
+	aws                  = "AWS"
+	gcp                  = "GCP"
 	gatewayConditionType = "SubmarinerGatewaysLabeled"
 )
 
@@ -42,6 +44,10 @@ var _ = Describe("Config Controller", func() {
 	testManagedClusterAddOn(t)
 })
 
+func labelGateway(node *corev1.Node, isGateway bool) {
+	node.Labels["submariner.io/gateway"] = strconv.FormatBool(isGateway)
+}
+
 func testWorkerNodeLabeling(t *configControllerTestDriver) {
 	When("no existing worker nodes are labeled as gateways", func() {
 		It("should label the desired number of gateway nodes", func() {
@@ -52,7 +58,7 @@ func testWorkerNodeLabeling(t *configControllerTestDriver) {
 
 	When("the desired number of gateway nodes are already", func() {
 		BeforeEach(func() {
-			t.nodes[0].Labels["submariner.io/gateway"] = "true"
+			labelGateway(t.nodes[0], true)
 		})
 
 		Context("partially labeled", func() {
@@ -104,7 +110,7 @@ func testWorkerNodeLabeling(t *configControllerTestDriver) {
 
 	When("the desired number of gateway nodes is increased", func() {
 		BeforeEach(func() {
-			t.nodes[0].Labels["submariner.io/gateway"] = "true"
+			labelGateway(t.nodes[0], true)
 			t.nodes[0].Labels["gateway.submariner.io/udp-port"] = strconv.Itoa(t.config.Spec.IPSecNATTPort)
 		})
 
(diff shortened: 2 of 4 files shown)


pkg/spoke/submarineragent/config_controller_test.go
```diff
@@ -52,7 +55,7 @@ func testWorkerNodeLabeling(t *configControllerTestDriver) {
 
 	When("the desired number of gateway nodes are already", func() {
 		BeforeEach(func() {
-			t.nodes[0].Labels["submariner.io/gateway"] = "true"
+			t.nodes[0].Labels["submariner.io/gateway"] = true_str
```
skitt, line 58: I don’t see the value here. The idea behind `goconst` is to ease the maintenance burden; a key indicator in my mind is whether or not the string being replaced can be given a name which reflects its _purpose_ rather than its value.

I’m curious why `goconst` doesn’t complain about the strings used as keys here, since they’re repeated too...

If we want to fix the repetitions in this set of functions, I would prefer extracting functions to set the labels:

```suggestion
			labelGateway(t.nodes[0])
```

pkg/spoke/submarineragent/config_controller_test.go
```diff
@@ -245,14 +248,14 @@ func testSubmarinerConfig(t *configControllerTestDriver) {
 		})
 	})
 
-	When("the SubmarinerConfig's Platform field is set to AWS", func() {
+	When("the SubmarinerConfig's Platform field is set to aWS", func() {
```
skitt, line 251: Overenthusiastic search and replace? ;-)

pkg/spoke/submarineragent/config_controller_test.go
dfarrell07: I made the Gateway-labeling function, and extracted all of those usages.

Goconst still thinks there are too many `true` values, however:

```
pkg/spoke/submarineragent/config_controller_test.go:48:41: string `true` has 4 occurrences, make it a constant (goconst)
	node.Labels["submariner.io/gateway"] = "true"
```

pkg/spoke/submarineragent/config_controller_test.go
```diff
@@ -351,7 +357,7 @@ func testSubmarinerConfig(t *configControllerTestDriver) {
 
 		Context("the SubmarinerConfig's Platform field is set to AWS", func() {
 			BeforeEach(func() {
-				t.config.Status.ManagedClusterInfo.Platform = "AWS"
+				t.config.Status.ManagedClusterInfo.Platform = aws
```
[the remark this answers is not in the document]
skitt: That would be good, yes!

.golangci.yml
```diff
@@ -96,3 +96,8 @@ issues:
     - linters:
         - lll
       path: "pkg/apis/submarinerconfig/v1alpha1/zz_generated.swagger_doc_generated.go"
+
+    # Replacing "true" with a constant doesn't make code more clear
+    - linters:
+        - goconst
+      text: "string `true` has"
```
skitt, line 103: Please fix the `"true"` mentions in `gateways_controller_test.go` instead of adding an exception.

pkg/spoke/submarineragent/config_controller_test.go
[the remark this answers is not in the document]
dfarrell07: As I tried to implement this various ways, they all seemed more complex than just letting the test code set the Platform inline using constants/"".

pkg/spoke/submarineragent/config_controller_test.go
```diff
@@ -42,6 +44,14 @@ var _ = Describe("Config Controller", func() {
 	testManagedClusterAddOn(t)
 })
 
+func labelGateway(node *corev1.Node, is_gateway bool) {
```
tpantelis, line 47:
```suggestion
func labelGateway(node *corev1.Node, isGateway bool) {
```

tpantelis (changes requested):

pkg/spoke/submarineragent/config_controller_test.go
```diff
@@ -42,6 +44,14 @@ var _ = Describe("Config Controller", func() {
 	testManagedClusterAddOn(t)
 })
 
+func labelGateway(node *corev1.Node, isGateway bool) {
+	if isGateway == true {
```
skitt, line 48: This is odd, I think one of the other linters should have flagged it — it should be `if isGateway {`. But see my suggestion...

pkg/spoke/submarineragent/config_controller_test.go
```diff
@@ -42,6 +44,14 @@ var _ = Describe("Config Controller", func() {
...
+		node.Labels["submariner.io/gateway"] = "true"
+	} else {
+		node.Labels["submariner.io/gateway"] = "false"
+	}
```
skitt, line 52:
```suggestion
	node.Labels["submariner.io/gateway"] = strconv.FormatBool(isGateway)
```

skitt (changes requested):

PR #183, opened 2021-10-08
Enable golangci-lint thelper linter, fix errors

dfarrell07:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -49,6 +49,7 @@ linters:
     - structcheck
     # - stylecheck  # instead of golint
     # - testpackage
+    - thelper
     - typecheck
     - unconvert
     # - unparam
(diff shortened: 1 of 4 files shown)


.golangci.yml
```diff
@@ -49,6 +49,7 @@ linters:
     - structcheck
     # - stylecheck  # instead of golint
     # - testpackage
+    - thelper
```
tpantelis, line 52: This is related to the Go testing framework however we're migrating all UTs to Ginkgo. Eventually all the functions to which you added `t.Helper() `will go away. Therefore I don't think we need that PR.

tpantelis (changes requested):

.golangci.yml
dfarrell07: Okay, closing. Thanks Tom!

PR #184, opened 2021-10-11
Add dependabot

Jaanki:


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+---
```
Jaanki, line 1: Starting with the most basic config. We can edit as and when required.

skitt: We already know that these will be necessary:

      # K8s and operator SDK, we need to handle these manually
      - dependency-name: github.com/operator-framework/*
      - dependency-name: k8s.io/*
      - dependency-name: sigs.k8s.io/*

tpantelis (approved):

PR #197, opened 2021-10-14
Enable golangci-lint gofumpt linter, fix errors

dfarrell07:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -53,7 +53,7 @@ linters:
     # - godox # TODO: Do we want this?
     # - goerr113 # TODO: Fix and enable
     - gofmt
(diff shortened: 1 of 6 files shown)


.golangci.yml
```diff
@@ -53,7 +53,7 @@ linters:
     # - godox # TODO: Do we want this?
     # - goerr113 # TODO: Fix and enable
     - gofmt
-    # - gofumpt # TODO: Do we want this?
+    - gofumpt # TODO: Do we want this?
```
tpantelis, line 56: This is billed as being a bit stricter than gofmt. Based on the changes in the source, it doesn't seem that compelling to me.

PR #198, opened 2021-10-14
Use latest Submariner in KIND setup

Jaanki:

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -206,7 +206,7 @@ spec:
 EOF
 
     submrepo="quay.io/submariner"
-    submver="0.9.1"
+    submver=0.11.0-rc1
     kubectl patch submarinerconfigs submariner -n ${cluster} --type "json" -p '[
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -206,7 +206,7 @@ spec:
 EOF
 
     submrepo="quay.io/submariner"
-    submver="0.9.1"
+    submver=$(curl -s https://github.com/submariner-io/submariner/releases/latest | grep -Po '(?<=tag/v)[^"]+')
```
skitt, line 209: Instead of scraping the HTML output, please use the relevant API; since we’re looking for image tags, use the Quay API:

```suggestion
    submver=$(curl https://quay.io/api/v1/repository/submariner/submariner-gateway/tag/ | jq -r '.tags[0].name')
```

skitt: Good points; I’m not sure we can resolve them satisfactorily. Since the addon has a different release schedule than Submariner itself, perhaps we should continue to hard-code this with whatever version the addon targets; so

```suggestion
    submver=0.11.0-rc1
```

until we get the 0.11.0 release (since the current development branch of the addon targets 0.11, as does the 2.5 branch).

PR #204, opened 2021-10-19
Document why golangci-lint forcetypeassert is disabled

dfarrell07:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -36,7 +36,8 @@ linters:
     # - exhaustivestruct # Not recommended for general use - meant to be used only for special cases
     - exportloopref
     # - forbidigo # We don't forbid any statements
-    # - forcetypeassert # TODO: Fix and enable
+    # - forcetypeassert # Most if not all unchecked type assertions would be the result of a programming
+    #     error so the reasonable recourse would be to panic anyway if checked so this doesn't seem useful.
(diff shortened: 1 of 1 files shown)


tpantelis:
I don't think there's any compelling reason to enable `forcetypeassert` - we have it disabled in other projects with the reason "_There are many unchecked type assertions that would be the result of a programming error so the reasonable recourse would be to panic anyway if checked so this doesn't seem useful_" which applies here as well. So let's just update the reason comment.

tpantelis (approved):

skitt:
Please avoid approving and merging PRs which only contain your own commits in future :-/. There wasn’t anything particularly urgent about this one...

tpantelis:
> Please avoid approving and merging PRs which only contain your own commits in future :-/. There wasn’t anything particularly urgent about this one...

This was originally @dfarrell07 commit and PR that was open for over 6 months.

skitt:
> > Please avoid approving and merging PRs which only contain your own commits in future :-/. There wasn’t anything particularly urgent about this one...
> 
> This was originally @dfarrell07 commit and PR that was open for over 6 months.

I know, but now it only has one commit, not written by Daniel.

skitt:
> yeah technically. But I just added doc text. We can revert it if you want.

No worries, the commit itself is fine, I would have approved it ;-).

PR #206, opened 2021-10-20
Enable golangci-lint nilerr linter, fix errors

tpantelis:


tpantelis:
CI job failed:

`Can't get enabled linters: unknown linters: 'nilerr', run 'golangci-lint linters' to see the list of supported linters`

So it seems the older **golangci-lint** version that prow uses doesn't contain the `nilerr` linter. Perhaps we can get it upgraded.  I'll leave this open for now.

PR #231, opened 2021-12-03
feature: install submariner broker and globalnet config on hub cluster

SataQiu:
Signed-off-by: SataQiu <shidaqiu2018@gmail.com>

/cc @sridhargaddam

### DIFF
--- a/pkg/hub/submarinerbroker/controller.go
+++ b/pkg/hub/submarinerbroker/controller.go
@@ -4,15 +4,19 @@ import (
 	"context"
 	"crypto/rand"
 	"embed"
+	"net"
+	"strconv"
+	"strings"
 
 	"github.com/open-cluster-management/submariner-addon/pkg/helpers"
 	brokerinfo "github.com/open-cluster-management/submariner-addon/pkg/hub/submarinerbrokerinfo"
 	"github.com/open-cluster-management/submariner-addon/pkg/resource"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/finalizer"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


SataQiu:
Hi @sridhargaddam 

Could you please take a look?

sridhargaddam:
> Hi @sridhargaddam
> 
> Could you please take a look?

Sure @SataQiu. I'll take a look at the PR tomorrow. Thanks.

sridhargaddam:
> Hi @sridhargaddam
> 
> Could you please take a look?

Hello @SataQiu, May I know how you are testing this feature? Are you using KIND or ACM GUI? 
If it's KIND, what options did you pass?

pkg/hub/submarinerbroker/controller.go
```diff
@@ -115,9 +131,57 @@ func (c *submarinerBrokerController) sync(ctx context.Context, syncCtx factory.S
...
 }
 
+func buildBrokerConfig(clusterSet *clusterv1beta1.ManagedClusterSet) (*brokerConfig, error) {
+	config := &brokerConfig{
+		SubmarinerNamespace:         brokerinfo.GenerateBrokerName(clusterSet.Name),
+		Components:                  []string{"service-discovery", "connectivity"},
+		DefaultCustomDomains:        nil,
+		GlobalnetCIDRRange:          "242.0.0.0/8",
+		DefaultGlobalnetClusterSize: 65536,
+		GlobalnetEnabled:            false,
```
sridhargaddam, line 141: I manually modified this to "true" and wanted to verify if globalnet is getting deployed with KIND Clusters (i.e., `make images`, `make clusters` and `make demo`). However, globalnet is not deployed.

Also, broker-globalnet-cm as well as broker objects are not created on the broker cluster.  

Can you please let us know how you are testing this change?

sridhargaddam (changes requested):

pkg/hub/submarinerbroker/controller.go
```diff
@@ -115,9 +131,57 @@ func (c *submarinerBrokerController) sync(ctx context.Context, syncCtx factory.S
...
+	}
+
+	if v, ok := clusterSet.Annotations[brokerComponentsAnnotationKey]; ok {
```
sridhargaddam, line 144: @skitt @tpantelis do you have any feedback on this approach of using annotations on the ManagedClusterSet object to enable certain features?

tpantelis: I don't know if this is a recommended approach and if other apps do it. For the submariner install on the managed clusters, we have a `SubmarinerConfig` CRD to communicate the user config so perhaps we should create an equivalent `BrokerConfig` CRD to be consistent.

SataQiu: I'm hesitant about this place, too. 
If `BrokerConfig` is a more acceptable solution, I can try it. @tpantelis @sridhargaddam

tpantelis: yeah let's try that. We don't own the `ManagedClusterSet` so our annotations could cause extraneous updates for others and possibly get overwritten.

Jaanki: Ideal way to create `Broker` on hub cluster is to install `submariner-operator` on the hub cluster and let the operator handle `broker` creation. `submariner` is created in the spoke in the similar fashion.

tpantelis: yeah it seems like we should be able to use the [Broker](https://github.com/submariner-io/submariner-operator/blob/devel/api/submariner/v1alpha1/submariner_types.go#L191) CRD and install the `submariner-operator`.

PR #235, opened 2021-12-07
Allow custom name for SubmarinerConfig

Jaanki:
This PR makes it possible to use a custom name for `SubmarinerConfig` rather than hard-coded `submariner` name.

Closes: https://github.com/open-cluster-management/submariner-addon/issues/221

### DIFF
--- a/pkg/hub/submarineragent/controller.go
+++ b/pkg/hub/submarineragent/controller.go
@@ -155,10 +155,6 @@ func NewSubmarinerAgentController(
 			// TODO: we may consider to use addon to set up the submariner env on the managed cluster instead of
 			// using manifestwork, one problem should be considered - how to get the cloud credentials
 			accessor, _ := meta.Accessor(obj)
-			if accessor.GetName() != helpers.SubmarinerConfigName {
-				return ""
-			}
-
 			return accessor.GetNamespace() + "/" + accessor.GetName()
 		}, configInformer.Informer()).
 		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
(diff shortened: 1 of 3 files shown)


tpantelis:
Is there a specific use case for customizing the CR name?

pkg/hub/submarineragent/controller.go
```diff
@@ -202,7 +198,7 @@ func (c *submarinerAgentController) sync(ctx context.Context, syncCtx factory.Sy
 			return err
 		}
 
-		config, err := c.configLister.SubmarinerConfigs(name).Get(helpers.SubmarinerConfigName)
+		config, err := c.configLister.SubmarinerConfigs(namespace).Get(name)
```
tpantelis, line 201: This is incorrect.  We get here b/c the `ManagedCluster`, `ManifestWork` or `ManagedClusterAddOn` changed, not the `SubmarinerConfig`, and `namespace` is empty so this lookup will always return `NotFound`.

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -86,6 +88,7 @@ func NewSubmarinerConfigController(input *SubmarinerConfigControllerInput) facto
 		clusterName:          input.ClusterName,
 		cloudProviderFactory: input.CloudProviderFactory,
 		onSyncDefer:          input.OnSyncDefer,
+		configName:           input.configName,
```
tpantelis, line 91: How is the caller to determine the `SubmarinerConfig` name on startup? Even if you query for it, it may not exist yet.

tpantelis (changes requested):

tpantelis:
This PR's been open for quite a while w/o any progression. As noted, allowing for a custom CR name is problematic and there doesn't seem to be a specific use case for it.

PR #245, opened 2022-01-12
Update Dockerfile to reference stolostron

tpantelis:
...instead of open-cluster-management.


skitt:
This is included in https://github.com/stolostron/submariner-addon/pull/244 which takes care of the org migration in general.

Issue #285, opened 2022-02-15
Addon changes to support globalnet configuration


sridhargaddam:
As part of supporting Globalnet in ACM, some of the changes related to the UI/Console will be handled via the following issue - https://github.com/stolostron/backlog/issues/19293

This issue is to track the associated changes in the submariner-addon component.

sridhargaddam:
Necessary changes are merged

PR #286, opened 2022-02-15
Support Globalnet deployment on managedClusters

sridhargaddam:
This PR implements the necessary support to enable Globalnet
deployments on managedClusters. It does this by parsing the
brokers.submariner.io object (named submariner-broker) in the
ManagedClusterSet broker namespace. If the brokers object has
Globalnet enabled, it allocates GlobalCIDR to each of the
managedClusters that are part of the ClusterSet.

The brokers.submariner.io object needs to created with
globalnetStatus set to false even for non-Globalnet deployments.

Fixes issue: https://github.com/stolostron/submariner-addon/issues/285


sridhargaddam:
CC: @tpantelis @skitt @vthapar

dfarrell07:
```
/go/src/github.com/stolostron/submariner-addon/vendor/git...m/submariner-io/submariner-operator/pkg/discovery/globalnet
pkg/hub/submarinerbrokerinfo/brokerinfo.go:16:2: cannot find package "." in:
	/go/src/github.com/stolostron/submariner-addon/vendor/github.com/submariner-io/submariner-operator/pkg/reporter
```

sridhargaddam:
I'm working on the Unit tests and linter issues.

dfarrell07:
Interesting, Sonarcloud is flagging something (maybe false-positive or minor): https://sonarcloud.io/project/security_hotspots?id=open-cluster-management_submariner-addon&pullRequest=286&resolved=false&types=SECURITY_HOTSPOT

sridhargaddam:
> Interesting, Sonarcloud is flagging something (maybe false-positive or minor): https://sonarcloud.io/project/security_hotspots?id=open-cluster-management_submariner-addon&pullRequest=286&resolved=false&types=SECURITY_HOTSPOT

Yeah, its the default GlobalnetCIDR. Is there a way to let Sonarcloud ignore this warning?

tpantelis:
> > Interesting, Sonarcloud is flagging something (maybe false-positive or minor): https://sonarcloud.io/project/security_hotspots?id=open-cluster-management_submariner-addon&pullRequest=286&resolved=false&types=SECURITY_HOTSPOT
> 
> Yeah, its the default GlobalnetCIDR. Is there a way to let Sonarcloud ignore this warning?

How about we reference it from submariner-operator? It's hardcoded [here](https://github.com/submariner-io/submariner-operator/blob/28296f046200f357294bbe067d9891a5e2f368a7/cmd/subctl/deploybroker.go#L72). We could add a constant, `DefaultCIDR`, in _pkg/discovery/globalnet/globalnet.go_ (also for the default `ClusterSize` as well).

pkg/hub/submarineragent/controller.go
```diff
@@ -497,6 +515,28 @@ func (c *submarinerAgentController) updateSubmarinerConfigStatus(ctx context.Con
 	return err
 }
 
+func (c *submarinerAgentController) updateManagedClusterAddOnStatus(ctx context.Context,
+	managedClusterAddon *addonv1alpha1.ManagedClusterAddOn, brokerNamespace string) error {
+	message := fmt.Sprintf("Waiting for brokers.submariner.io object named submariner to be created in %q namespace", brokerNamespace)
```
tpantelis, line 520:
```suggestion
	message := fmt.Sprintf("Waiting for brokers.submariner.io object named %q to be created in %q namespace", brokerObjectName, brokerNamespace)
```

pkg/hub/submarineragent/controller.go
```diff
@@ -497,6 +515,28 @@ func (c *submarinerAgentController) updateSubmarinerConfigStatus(ctx context.Con
...
+	condition := metav1.Condition{
+		Type:    BrokerCfgMissing,
+		Status:  metav1.ConditionFalse,
+		Reason:  "SubmarinerBrokerConfigMissing",
+		Message: message,
+	}
```
tpantelis, line 526: The `Status` should be `metav1.ConditionTrue` since we're asserting that the broker config is missing. Also, we should assert the opposite `Status` once the broker config is available.

However, I think the condition should reflect the status of the globalnet config. So define the type as "_GlobalnetConfigApplied_". In this case, it would have `metav1.ConditionFalse` with `Reason` "_BrokerConfigMissing_". When the globalnet config is applied, then update `Status`  to `metav1.ConditionTrue`.

tpantelis (changes requested):

pkg/hub/submarineragent/controller.go
```diff
@@ -58,6 +62,10 @@ const (
 	submarinerConfigFinalizer    = "submarineraddon.open-cluster-management.io/config-cleanup"
 	agentRBACFile                = "manifests/rbac/operatorgroup-aggregate-clusterrole.yaml"
 	submarinerCRFile             = "manifests/operator/submariner.io-submariners-cr.yaml"
+	BrokerCfgMissing             = "SubmarinerBrokerConfigMissing"
+	brokerObjectName             = "submariner"
+	defaultGNClusterSize         = 65535
```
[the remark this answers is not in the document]
sridhargaddam: It will be rounded to the same value, but I'll update it to keep it consistent.

sridhargaddam (commented):
Thanks for reviewing @tpantelis

sridhargaddam:
> How about we reference it from **submariner-operator**? It's hardcoded [here](https://github.com/submariner-io/submariner-operator/blob/28296f046200f357294bbe067d9891a5e2f368a7/cmd/subctl/deploybroker.go#L72). We could add a constant, `DefaultCIDR`, in _pkg/discovery/globalnet/globalnet.go_ (also for the default `ClusterSize` as well).

Good idea. Done - https://github.com/submariner-io/submariner-operator/pull/1858

pkg/hub/submarineragent/controller.go
tpantelis: But this condition only indicates that it's missing w/o indicating any ramifications. I think it makes more sense to assert a condition indicating what you're trying to actually do, in this case apply the globalnet config. A missing Broker object is one reason the condition isn't satisfied - if an error occurs we could also indicate that in the reason/message. Also when it does get successfully applied, the message can include the info that was applied, ie the CIDR etc.

In the future, if we apply other config from the Broker object, we can assert separate conditions appropriately ("_XYZApplied_").

sridhargaddam:
Sonar Cloud Quality Gate Passed :)

pkg/hub/submarineragent/controller.go
sridhargaddam: @tpantelis is it okay to address it in a separate PR and get the current one merged?

tpantelis (approved):

PR #291, opened 2022-02-23
Update the catalog manifests

skitt:
This updates the catalog manifests with recent changes to the CRDs,
RBAC etc.


dfarrell07:
@sridhargaddam You have permission to use /allow (+2) here, if you'd like. https://github.com/stolostron/submariner-addon/blob/main/OWNERS#L17

tpantelis (approved):

PR #295, opened 2022-02-24
Run diagnose Job

vthapar:
Diagnose implementation


pkg/spoke/submarineragent/diagnose_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		namespace string) error {
+	var backOffLimit int32 = 0
+	jobSpec := &batchv1.Job{
+		ObjectMeta : metav1.ObjectMeta{
+			Name: jobName,
+			Namespace: namespace,
+		},
+		Spec: batchv1.JobSpec{
+			Template: v1.PodTemplateSpec{
+				Spec: v1.PodSpec{
```
tpantelis, line 171: Don't you also need the pod template labels, ie

```
ObjectMeta: metav1.ObjectMeta{
    Labels: map[string]string{jobLabel: jobName},
},
```

pkg/spoke/submarineragent/diagnose_controller.go
```diff
@@ -0,0 +1,219 @@
...
+
+			return metaObj.GetName() == constants.SubmarinerConfigName
+		}, input.ConfigInformer.Informer()).
+		WithFilteredEventsInformers(func(obj interface{}) bool {
+			metaObj := obj.(metav1.Object)
+			if jobLabelValue, has := metaObj.GetLabels()[jobLabel]; has {
+				if jobLabelValue == jobName {
+					return true
+				}
+			}
```
tpantelis, line 88: It seems expensive to watch all pods. Perhaps we can watch Jobs instead?

tpantelis (changes requested):

pkg/spoke/submarineragent/diagnose_controller.go
vthapar: K8s adds the label for all pods created by jobs.

pkg/spoke/submarineragent/diagnose_controller.go
vthapar: Yeah, that was my concern too, how does this filtered informer work? If it is implemented in backend as a single informer for all controllers and just notifies us all of events, that should be fine. Jobs won't work as we need the podname and can't be sure if pod is available by the time we get Job notification. Is there a way to watch for pods within a specific namespace only? Alternately, would it be okay to poll for pod creation?

tpantelis: I think you can watch for the Job and query for the associated pod as you're doing. If it doesn't exist then requeue. Or just wait for the next Job notification. I assume you want to update the status condition when the Job completes? If so then perhaps check the Job status rather the querying the pod.

vthapar: Getting back to it with 2.5 delayed.

@tpantelis We don't even need to wait for Job completion, just need the name of the pod so in UI we can link to that pod's logs. UI can stream logs in realtime, which works even better as user can see the progress of diagnose from the logs.

pkg/spoke/submarineragent/diagnose_controller.go
```diff
@@ -0,0 +1,214 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/stolostron/submariner-addon/pkg/apis/submarinerdiagnoseconfig"
+	diagnosev1alpha1 "github.com/stolostron/submariner-addon/pkg/apis/submarinerdiagnoseconfig/v1alpha1"
+	diagnoseclient "github.com/stolostron/submariner-addon/pkg/client/submarinerdiagnoseconfig/clientset/versioned"
+	diagnoseinformer "github.com/stolostron/submariner-addon/pkg/client/submarinerdiagnoseconfig/informers/externalversions/submarinerdiagnoseconfig/v1alpha1" //nolint:lll // Ignore line length check
```
skitt, line 12:
```suggestion
	diagnoseinformer "github.com/stolostron/submariner-addon/pkg/client/submarinerdiagnoseconfig/informers/externalversions/submarinerdiagnoseconfig/v1alpha1" //nolint
```

skitt (changes requested):

pkg/spoke/submarineragent/diagnose_controller.go
tpantelis: Revisiting this.... I'm still wary of the overhead of watching all pods. I still think it would be better to watch for the `Job` and query for the associated pod as outlined above.

PR #298, opened 2022-03-02
Add RHOS Support

vthapar:


nyechiel:
We want this in 2.5, need a proper environment to test.

aswinsuryan:
I tested with ACM2.4 and ensured that ports are open and gateway is tagged, but not the end to end flow since the clusters had overlapping CIDRS and global net is not available in acm2.4.

aswinsuryan (approved):

pkg/cloud/rhos/rhos.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"github.com/submariner-io/cloud-prepare/pkg/api"
+	cloudpreparerhos "github.com/submariner-io/cloud-prepare/pkg/rhos"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	cloudsYAMLName = "clouds.yaml"
+	cloudName      = "cloud"
+	gwInstanceType = "PnTAE.CPU_4_Memory_8192_Disk_50"
```
sridhargaddam, line 33: I have a feeling that this is not a common flavor that is available across all OCP deployments on OSP, but since @aswinsuryan mentioned earlier that it is what is selected with OCP, we can move ahead. The reason I'm emphasising this again is because currently we are not showing the flavor option to the user, so in the absence of this flavor, cloud-prepare would run into issues.

pkg/cloud/rhos/rhos.go
```diff
@@ -0,0 +1,204 @@
...
+		reporter:             reporter.NewEventRecorderWrapper("RHOSCloudProvider", eventRecorder),
+		nattDiscoveryPort:    int64(nattDiscoveryPort),
+		gateways:             gateways,
+	}, nil
+}
+
+// PrepareSubmarinerClusterEnv prepares submariner cluster environment on RHOS
+// The below tasks will be executed
+// 1. create the inbound and outbound firewall rules for submariner, below ports will be opened
+//    - IPsec IKE port (by default 500/UDP)
```
sridhargaddam, line 114:
```suggestion

```
Please remove this line as we are not opening the 500/UDP Port. In its place, you can add ESP, AH ports

pkg/cloud/rhos/rhos.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+
+	r.reporter.Succeeded("The Submariner cluster environment has been set up on RHOS")
+	return nil
+}
+
+// CleanUpSubmarinerClusterEnv clean up submariner cluster environment on RHOS after the SubmarinerConfig was deleted
+// 1. delete the inbound and outbound firewall rules to close submariner ports
+// 2. delete the inbound and outbound firewall rules to close submariner metrics port
```
sridhargaddam, line 150: This comment needs to be appropriately updated.

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -249,9 +250,10 @@ func (c *submarinerConfigController) cleanupClusterEnvironment(ctx context.Conte
 		}
 
 		return errors.WithMessagef(err, "Failed to clean up the submariner cluster environment")
-	}
-
-	if config.Status.ManagedClusterInfo.Platform != "AWS" {
+	case "AWS":
+		// Nothing to do
```
sridhargaddam, line 254: You can add a comment that cloud-prepare for AWS is handled on Hub, so nothing to be done here (aka Spoke)

sridhargaddam (changes requested):

pkg/cloud/rhos/rhos.go
vthapar: @sridhargaddam To deploy submariner on RHOS through addon, we ned to manually create submarinerconfig. So user can always specify custom flavor when creating Submarinerconfig. By the time we add RHOS to UI, option will be made available.

sridhargaddam: I thought the SubmarinerConfig will be automatically created when user tries to install Submariner Addon on the ManagedCluster running on RHOS. However, we can always **update** it, if required. Is this not the case?

pkg/cloud/rhos/rhos.go
```diff
@@ -0,0 +1,204 @@
...
+	}, nil
+}
+
+// PrepareSubmarinerClusterEnv prepares submariner cluster environment on RHOS
+// The below tasks will be executed
+// 1. create the inbound and outbound firewall rules for submariner, below ports will be opened
+//    - NAT traversal port (by default 4500/UDP)
+//    - 4800/UDP port to encapsulate Pod traffic from worker and master nodes to the Submariner Gateway nodes
+//    - ESP & AH protocols for private-ip to private-ip gateway communications
+// 2. create the inbound and outbound firewall rules to open 8080/TCP port to export metrics service from the Submariner gateway
```
sridhargaddam, line 117:
```suggestion
// 2. create the inbound and outbound firewall rules to open 8080/TCP, 8081/TCP ports to export metrics service from the Submariner gateway
```

pkg/cloud/rhos/rhos.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return err
+	}
+
+	r.reporter.Succeeded("The Submariner cluster environment has been set up on RHOS")
+	return nil
+}
+
+// CleanUpSubmarinerClusterEnv clean up submariner cluster environment on RHOS after the SubmarinerConfig was deleted
+// 1. deleted any dedicated gateways that were previously deployed.
```
sridhargaddam, line 148:
```suggestion
// 1. delete any dedicated gateways that were previously deployed.
```

pkg/cloud/rhos/rhos.go
vthapar: It is done by UI for clouds that are supported through UI. for RHOS, vSphere etc. It will at most create ManagedClusterAddon object in that cluster's namespace on hub. Note that this is the UI use case where you add non supported cluster to a clusterset with submariner addon already installed.

sridhargaddam (approved):

PR #299, opened 2022-03-02
Remove dedicated g/w node flag

aswinsuryan:
* Remove dedicated g/w node flag from SubmarinerConfig CR


skitt:
This is removing a field that was added recently, so it won’t break upgrades.

/approve

vthapar (approved):

Issue #304, opened 2022-03-03
Upgrade the addon to Submariner 0.12.0-rc0


skitt:
This involves a number of changes.

- [x] Upgrade the dependencies in `go.mod` and deal with the fallout.
- [x] Ensure that 0.12.0-rc0 images are available in the appropriate repository.
- [x] Change the default deployed version in `pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml`.
- [x] Change the deployed version in `scripts/deploy.sh`.

Jaanki:
Updated deploy script to use subm v0.12.0-rc0 in https://github.com/stolostron/submariner-addon/pull/307

Jaanki:
0.12.0-rc0 images are available in quay.io/submariner as well.

PR #306, opened 2022-03-03
Install version 0.12.0 of the operator 2.5 release

MaxBab:
Should fix - https://bugzilla.redhat.com/show_bug.cgi?id=2060469


skitt:
This is being tracked in https://github.com/stolostron/submariner-addon/issues/304. We don’t have a v0.12.0 release yet.

Issue #309, opened 2022-03-09
Globalnet metrics not shown up with OCP Deployments.


sridhargaddam:
After installing Submariner addon on a managed cluster, when I look at OCP metrics dashboard, I dont find any of the Globalnet metrics in the list.

sridhargaddam:
Upon debugging the issue, the problem seems to be related to firewall ports not being configured on the underlay. Currently as part of cloud-prepare we are only opening TCP/8080 metrics port which is for submariner gateway pod. When I manually opened TCP/8081 (which is where Globalnet metrics are exposed), I can see the metrics in OCP metrics dashboard.

![Screenshot from 2022-03-09 13-52-30](https://user-images.githubusercontent.com/6574656/157409391-58b520b8-a20f-4276-b206-4f4102573b94.png)

PR #316, opened 2022-03-11
Allow Globalnet metrics port while preparing cloud

sridhargaddam:
Submariner Globalnet pod runs with HostNetworking enabled and the
metrics are exported on TCP/8081 port. Currently, this port is
not configured while preparing the cloud infrastructure, because
of this Globalnet metrics are not seen in OCP dashboard. This PR
fixes the issue.

Fixes: https://github.com/stolostron/submariner-addon/issues/309

### DIFF
--- a/pkg/cloud/aws/aws.go
+++ b/pkg/cloud/aws/aws.go
@@ -124,7 +124,8 @@ func (a *awsProvider) PrepareSubmarinerClusterEnv() error {
 	if err := a.cloudPrepare.PrepareForSubmariner(cpapi.PrepareForSubmarinerInput{
 		InternalPorts: []cpapi.PortSpec{
 			{Port: constants.SubmarinerRoutePort, Protocol: "udp"},
-			{Port: constants.SubmarinerMetricsPort, Protocol: "tcp"},
+			{Port: constants.SubmarinerGatewayMetricsPort, Protocol: "tcp"},
+			{Port: constants.SubmarinerGlobalnetMetricsPort, Protocol: "tcp"},
 		},
 	}, a.reporter); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


sridhargaddam:
CC @skitt @aswinsuryan @vthapar @nyechiel 

In this PR I'm opening the Globalnet Metrics port unconditionally. Ideally we should open it only when Globalnet is enabled.
However, in order to do that it requires many changes because 
1. The cloud preparation for clouds like AWS is done on Hub whereas its done on Spoke for GCP/RHOS.
2. The cloud preparation APIs in addon currently have an hard-coded list of internal ports. In order to support GlobalnetMetricsPort we will have to modify the APIs.
3. Currently the Spoke cluster has access to SubmarinerConfig on the Hub, but Globalnet is a ClusterSet wide configuration and users are not really expected to supply GlobalCIDR (unless explicitly overriding) in the submarinerConfig. Submariner-addon code will look at broker configuration in the managedClusterSetBrokre Namespace and will allocate a unique GlobalCIDR for each cluster which is not stored in the submarinerConfig. If Spoke has to read the broker config or local submariner config, it requires many changes.

Since we are very close to a release, IMHO its not advisable to make major changes. Also, the metrics port is an **internal** port which will only be allowed within the local cluster and NOT to outside of the cluster. 

In the next release, IMO we should improve the cloud-prepare stuff to either Hub/Spoke and once its done, we can revisit this use-case to enable the Globalnet Metrics port accordingly.

skitt:
I agree, let’s keep this simple. Since it’s temporary, I won’t ask for a `metricsPorts` slice instead of the two explicit fields in `gcpProvider` ;-).

/approve

(Leaving lgtm to a second reviewer.)

PR #327, opened 2022-03-22
Fix platform name for Openstack clouds

aswinsuryan:
Fix platform name for Openstack clouds

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -77,7 +77,7 @@ func (f *providerFactory) Get(managedClusterInfo configv1alpha1.ManagedClusterIn
 		return gcp.NewGCPProvider(f.restMapper, f.kubeClient, f.dynamicClient, f.hubKubeClient, eventsRecorder,
 			region, infraID, clusterName, config.Spec.CredentialsSecret.Name,
 			config.Spec.GatewayConfig.GCP.InstanceType, config.Spec.IPSecNATTPort, config.Spec.NATTDiscoveryPort, config.Spec.Gateways)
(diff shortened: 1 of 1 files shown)


vthapar (approved):

pkg/cloud/cloud.go
```diff
@@ -77,7 +77,7 @@ func (f *providerFactory) Get(managedClusterInfo configv1alpha1.ManagedClusterIn
 		return gcp.NewGCPProvider(f.restMapper, f.kubeClient, f.dynamicClient, f.hubKubeClient, eventsRecorder,
 			region, infraID, clusterName, config.Spec.CredentialsSecret.Name,
 			config.Spec.GatewayConfig.GCP.InstanceType, config.Spec.IPSecNATTPort, config.Spec.NATTDiscoveryPort, config.Spec.Gateways)
-	case "Openstack":
+	case "OpenStack":
```
sridhargaddam, line 80: This looks like a fundamental bug. Just curious, is this the first attempt where we are validating the end-to-end flow with the RHOS PRs?

sridhargaddam:
/lgtm

Can we please ensure that we do end-to-end testing of this feature on RHOS with all the associated PRs and identify any missing elements. Since we do not have a CI job to validate this its important that we test this properly during manual validation.

pkg/cloud/cloud.go
aswinsuryan: This was a typo that occurred in the original PR. As I was testing the changes with my local changes it had the correct platform name, but the PR did to have it and was missed in the review.

PR #355, opened 2022-05-02
Rebuild images for CVE updates

dfarrell07:


dfarrell07:
It doesn't seem like this way of making a noop works with OWNERS/Prow either.

PR #373, opened 2022-05-18
Add image override for networkPluginSyncer

vthapar:
This adds option to override NetworkPluginSyncer image in `ImagePullspecs`. Submariner supports this override but option was missing from addon.


Jaanki (approved):

skitt:
@nyechiel I suggested Vishal issue the cherrypick command when we discussed this yesterday, but that was before the conclusion in the BZ that this was 2.6 material since it’s only needed for OVN. I’ll cancel the cherrypick PR once it’s opened if this doesn’t work:

/uncherrypick release-2.5

Issue #218, opened 2021-11-02
Move AWS cloud preparation to the spoke


tpantelis:
GCP cloud preparation is done in the spoke so, for consistency, simplicity, and maintainability, we should perform all cloud preparation in the spoke. This would make the controllers platform agnostic and put all platform decision logic in the cloud provider factory.

cmwylie19:
initial investigation:
- Are there any issues moving this (AWS) to the spoke?

nyechiel:
@skitt @tpantelis while still doable, I wonder if it's not too late/risky for 2.7. Wdyt?

skitt:
I agree, it’s too late for this. We can move it to 2.8.

tpantelis:
> @skitt @tpantelis while still doable, I wonder if it's not too late/risky for 2.7. Wdyt?

I think it's risky no matter what release it goes into 😄 It looks like code freeze is Dec 29th but with the holidays it's about 3 weeks including testing in a real AWS env (which I'd need some help/guidance with).

nyechiel:
> > @skitt @tpantelis while still doable, I wonder if it's not too late/risky for 2.7. Wdyt?
> 
> I think it's risky no matter what release it goes into smile It looks like code freeze is Dec 29th but with the holidays it's about 3 weeks including testing in a real AWS env (which I'd need some help/guidance with).

Let's move out from 2.7 for now. We can revisit for 2.8.

PR #381, opened 2022-06-02
Add support for Azure cloud preperation

aswinsuryan:


nyechiel:
@aswinsuryan Have you managed to give it a try with OCP/ACM? Any estimation on when this can be moved out of draft?

aswinsuryan:
> @aswinsuryan Have you managed to give it a try with OCP/ACM? Any estimation on when this can be moved out of draft?

It is now ready for review.

pkg/hub/submarineragent/controller.go
```diff
@@ -785,3 +785,11 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(brokerNamespace
...
 	return nil
 }
+
+func isSpokePrepared(cloudName string) bool {
+	if cloudName == "GCP" || cloudName == "OpenStack" || cloudName == "Azure" {
+		return true
+	}
+
+	return false
+}
```
skitt, line 795: :+1:

pkg/cloud/azure/azure.go
```diff
@@ -0,0 +1,211 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+)
+
+const (
+	servicePrincipalJSON = "osServicePrincipal.json"
+	gwInstanceType       = "Standard_D4s_v3"
```
sridhargaddam, line 29: We are using a different instance type here compared to the instance type of `Standard_F4s_v2` used as part of  `subctl cloud prepare azure`. Is this a typo or intentional?

pkg/cloud/azure/azure.go
```diff
@@ -0,0 +1,211 @@
...
+		BaseGroupName:  infraID + "-rg",
+		Authorizer:     authorizer,
+		K8sClient:      k8sClient,
+	}
+
+	azureCloud := azure.NewCloud(&cloudInfo)
+
+	gwDeployer, err := azure.NewOcpGatewayDeployer(&cloudInfo, azureCloud, msDeployer, instanceType, true)
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to initialize a GatewayDeployer config")
```
sridhargaddam, line 100:
```suggestion
		return nil, errors.Wrap(err, "failed to initialise a GatewayDeployer config")
```

pkg/cloud/azure/azure.go
```diff
@@ -0,0 +1,211 @@
...
+			{Port: r.nattPort, Protocol: "udp"},
+			{Port: uint16(r.nattDiscoveryPort), Protocol: "udp"},
+			{Port: 0, Protocol: "esp"},
+			{Port: 0, Protocol: "ah"},
+		},
+		Gateways: r.gateways,
+	}, r.reporter); err != nil {
+		return err
+	}
+
```
sridhargaddam, line 139: Please add a "TODO:" note here saying that for OVN Clusters, we have to avoid opening the UDP/4800 ports. @vthapar when working on the necessary OVN PR can modify this part as well.

pkg/cloud/azure/azure.go
```diff
@@ -0,0 +1,211 @@
...
+	if !ok {
+		return "", errors.New("servicePrincipalJSON is not found in the credentials")
+	}
+
+	err = json.Unmarshal(servicePrincipalJSON, &authInfo)
+	if err != nil {
+		return "", errors.Wrap(err, "error unmarshalling servicePrincipalJSON")
+	}
+
+	if err = os.Setenv("AZURE_CLIENT_ID", authInfo.ClientId); err != nil {
```
sridhargaddam, line 198: Is it safe to assume that we can update the environment variable inside the container? Are there any alternate way to pass this value? 

Currently we are executing the cloud prepare in the context of spoke, but when we move this in the context of Hub (in future PR) it might lead to issues as the environment variable will be global in the container image.

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -573,3 +573,11 @@ func successCondition(gatewayNames []string) metav1.Condition {
...
 	}
 }
+
+func isSpokePrepared(cloudName string) bool {
+	if cloudName == "GCP" || cloudName == "OpenStack" || cloudName == "Azure" {
+		return true
+	}
+
+	return false
+}
```
sridhargaddam, line 583: This is duplicated, please create a common method and use it on the hub/spoke.

sridhargaddam (changes requested):
Thanks for the PR @aswinsuryan

pkg/cloud/azure/azure.go
aswinsuryan: I was not able to find an alternate way to do this. subctl also uses a similar way. If needs to be moved to hub we need to explore possible options then. It is just to create the authorizer, once created we can use the authorizer to authenticate.

pkg/cloud/azure/azure.go
aswinsuryan: was a an error when took from a different deployment will make it same as subctl

pkg/cloud/azure/azure.go
aswinsuryan: Done

pkg/spoke/submarineragent/config_controller.go
tpantelis: Once we move AWS to the spoke (see https://github.com/stolostron/submariner-addon/issues/218) then this becomes unnecessary so I don't think it's worth creating a common function for this.

sridhargaddam: Okay, but what about the default case? Do we not have to distinguish between the clouds where we have automation vs clouds like VmWare/BareMetal etc which requires manual preparation?

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -573,3 +573,11 @@ func successCondition(gatewayNames []string) metav1.Condition {
 			strings.Join(gatewayNames, ",")),
 	}
 }
+
```
sridhargaddam, line 576: The comment below was marked as resolved and I'm not getting option to unmark it. One follow up question.

What about the default case? Do we not have to distinguish between the clouds where we have automation (like AWS/GCP/RHOS/Azure) vs clouds like VmWare/BareMetal etc which requires manual preparation?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/stolostron/submariner-addon
 go 1.18
 
 require (
+	github.com/Azure/go-autorest/autorest/azure/auth v0.5.7
```
sridhargaddam, line 6: Small typo in the commit message. If you happen to respin the PR
preperation -> preparation

pkg/hub/submarineragent/controller.go
```diff
@@ -785,3 +785,11 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(brokerNamespace
 
 	return nil
 }
+
+func isSpokePrepared(cloudName string) bool {
+	if cloudName == "GCP" || cloudName == "OpenStack" || cloudName == "Azure" {
+		return true
+	}
+
+	return false
```
tpantelis, line 794: I believe only AWS is done on the hub, if so:

```suggestion
	return cloudName != "AWS"
```

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -243,7 +243,7 @@ func (c *submarinerConfigController) cleanupClusterEnvironment(ctx context.Conte
 	recorder events.Recorder,
 ) error {
 	switch config.Status.ManagedClusterInfo.Platform {
-	case "GCP", "Openstack":
+	case "GCP", "Openstack", "Azure":
```
tpantelis, line 246: Why don't we use `isSpokePrepared` here:

```Go
if isSpokePrepared(config.Status.ManagedClusterInfo.Platform) {
    cloudProvider, err := c.cloudProviderFactory.Get(config.Status.ManagedClusterInfo, config, recorder)
   
    return ...
}

return errors.WithMessagef(c.removeAllGateways(ctx), "Failed to unlabel the gateway nodes")
```

tpantelis (changes requested):

pkg/spoke/submarineragent/config_controller.go
tpantelis: Yes but only here - once AWS is moved here then the hub controller won't do any cloud prepare.

sridhargaddam: My question is not about Hub/Spoke, but its about - dont we want to distinguish between clouds which have automation (like AWS/GCP/Azure/RHOS) vs the remaining?

pkg/hub/submarineragent/controller.go
aswinsuryan: But since any thing other that does not use submariner cloud prepare (like vmware ) need to take a different path,

pkg/spoke/submarineragent/config_controller.go
aswinsuryan: Done.

pkg/hub/submarineragent/controller.go
tpantelis: Not sure I follow.

aswinsuryan: I mean to say there are three cases
if it is AWS it will take this path 
https://github.com/stolostron/submariner-addon/blob/main/pkg/spoke/submarineragent/config_controller.go#L174

if it s GCP, Azure  or Openstack  it takes this path
https://github.com/stolostron/submariner-addon/blob/main/pkg/spoke/submarineragent/config_controller.go#L180

if it is VMware or any other cloud it takes this path
https://github.com/stolostron/submariner-addon/blob/main/pkg/spoke/submarineragent/config_controller.go#L185

is it is not AWS we need to check if it is case 2 or case3

aswinsuryan: ah oh ok got it it was my bad. I was just thinking it is config controller in spoke

aswinsuryan: Done.

pkg/cloud/azure/azure.go
```diff
@@ -0,0 +1,212 @@
...
+		nattDiscoveryPort = constants.SubmarinerNatTDiscoveryPort
+	}
+
+	subscriptionID, err := initializeFromAuthFile(hubKubeClient, clusterName, credentialsSecretName)
+
+	authorizer, err := auth.NewAuthorizerFromEnvironment()
+
+	if err != nil {
+		klog.Errorf("unable to create the Azure  authorizer :%v", err)
+		return nil, err
```
tpantelis, line 80:
```suggestion
		return nil, errors.Wrap(err, "unable to create the Azure authorizer")
```

pkg/spoke/submarineragent/config_controller.go
sridhargaddam: The isSpokePrepared API was exactly the same in the initial PR (so i added a comment to de-duplicate it) and was later modified in subsequent PRs. It looks fine now.

sridhargaddam (approved):

pkg/hub/submarineragent/controller.go
```diff
@@ -374,7 +374,7 @@ func (c *submarinerAgentController) syncSubmarinerConfig(ctx context.Context,
 	// prepare submariner cluster environment
 	errs := []error{}
 	var condition *metav1.Condition
-	if managedClusterInfo.Platform != "GCP" && managedClusterInfo.Platform != "OpenStack" {
+	if !isSpokePrepared(managedClusterInfo.Platform) {
```
tpantelis, line 377: On second thought wrt `isSpokePrepared` as I suggested, for a provider not covered by cloud prepare, we won't set the `SubmarinerConfigConditionEnvPrepared` condition. By keeping the checks as they were then here we'll cover AWS and any other not handled by the spoke. For the latter, cloud prepare is a no-op but we will still set the condition here (the spoke will not). Next release we should move it all into the spoke as it's currently a bit confusing.

tpantelis (changes requested):

tpantelis:
@sridhargaddam This was prematurely merged. I still have outstandiing comments.

sridhargaddam:
> @sridhargaddam This was prematurely merged. I still have outstandiing comments.

Oops, I should have used "/lgtm" i guess. @aswinsuryan can you handle the review comments from @tpantelis in a separate PR?

tpantelis:
> > @sridhargaddam This was prematurely merged. I still have outstandiing comments.
> 
> Oops, I should have used "/lgtm". @aswinsuryan can you handle the review comments from @tpantelis in a separate PR?

I think we should move the default handling into the spoke - I can look into that. Ideally it would be great to  move AWS into the spoke as well got 2.6 but I that might be too tight.

nyechiel:
> > > @sridhargaddam This was prematurely merged. I still have outstandiing comments.
> > 
> > 
> > Oops, I should have used "/lgtm". @aswinsuryan can you handle the review comments from @tpantelis in a separate PR?
> 
> I think we should move the default handling into the spoke - I can look into that. Ideally it would be great to move AWS into the spoke as well for 2.6 but I that might be too tight.

@tpantelis @aswinsuryan what's the plan here? are we expecting further changes for 2.6.0, or should we let QE start testing?

tpantelis:
> @tpantelis @aswinsuryan what's the plan here? are we expecting further changes for 2.6.0, or should we let QE start testing?

I'm working on further changes although we don't really need them for 2.6. I thought we had until Jul 27th?

nyechiel:
> > @tpantelis @aswinsuryan what's the plan here? are we expecting further changes for 2.6.0, or should we let QE start testing?
> 
> I'm working on further changes although we don't really need them for 2.6. I thought we had until Jul 27th?

Indeed. We just need to be clear with QE on where things stand. I believe @MaxBab is planning to incorporate these changes into his test pipeline (replacing existing manual cloud-preparations steps).

aswinsuryan:
> > > @tpantelis @aswinsuryan what's the plan here? are we expecting further changes for 2.6.0, or should we let QE start testing?
> > 
> > 
> > I'm working on further changes although we don't really need them for 2.6. I thought we had until Jul 27th?
> 
> Indeed. We just need to be clear with QE on where things stand. I believe @MaxBab is planning to incorporate these changes into his test pipeline (replacing existing manual cloud-preparations steps).

The changes @tpantelis working on does not effect the UX. It is technical debt.

PR #400, opened 2022-06-20
Trigger a rebuild

dfarrell07:
These two commits undo each other, but hopefully the will trigger a rebuild.

We tried an empty commit in #399 but that didn't work.


dfarrell07:
>    	github.com/stolostron/submariner-addon/pkg/version	[no test files]

Weird.

dfarrell07:
And now it's passing

skitt:
Ah, an empty PR can’t be approved.

Issue #386, opened 2022-06-08
cloud prepare for AWS is invoked periodically


sridhargaddam:
Currently, looking at the submariner-addon code, it looks like we are [periodically invoking cloud-prepare for AWS](https://github.com/stolostron/submariner-addon/blob/3a399d447a1e086c53cd2023462e81a87536f8cb/pkg/hub/submarineragent/controller.go#L365) clusters eventhough the cloud is already prepared. This is not causing an issue because the cloud prepare apis are idempotent. However, this can be optimised by check the submarinerConfig status and invoking cloud prepare APIs only when its not yet done - [similar to the way we do for GCP/OSP clusters](https://github.com/stolostron/submariner-addon/blob/9e00467d29a7731afe830141208e5844deb30565/pkg/spoke/submarineragent/config_controller.go#L199).

mkolesnik:
This seems like the behavior I observed in https://bugzilla.redhat.com/show_bug.cgi?id=2093990#c6

Though it seems it causes gateway nodes to be destroyed and redeployed each time.

PR #411, opened 2022-07-04
Skip syncing unchanged submariner configs

mkolesnik:
If the actual configuration didn't change (the spec) and the config
isn't being deleted, then skip syncing it to avoid unnecessarily calling
cloud prepare all the time.

Resolves: #386

Signed-off-by: Mike Kolesnik <mkolesni@redhat.com>

### DIFF
--- a/pkg/hub/submarineragent/controller.go
+++ b/pkg/hub/submarineragent/controller.go
@@ -5,6 +5,7 @@ import (
 	"embed"
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"github.com/ghodss/yaml"
 	"github.com/openshift/library-go/pkg/assets"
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):

mkolesnik:
> These changes alleviate the issue but long term I think we should consider refactoring to use the admiral resource syncer which allows for granular updates.

Yes it seems that the same syncer is triggered by multiple events on different resources, so its necessary to understand the chain of events here before refactoring.

Can the admiral syncer help with not triggering sync in case spec didn't change?

tpantelis:
> Yes it seems that the same syncer is triggered by multiple events on different resources, so its necessary to understand the chain of events here before refactoring.
> 
> Can the admiral syncer help with not triggering sync in case spec didn't change?

Yes. A lower-level informer provides access to the previous instance on an update and the admiral syncer utilizes that to allow the user to compare.

PR #453, opened 2022-07-22
Move default provider handling to the spoke

tpantelis:
...specifically the case where there's no cloud provider implementation for a platform. We still want to set the status
condition indicating the environment was prepared. Previously this was done on the hub but it should be done on the spoke as only AWS is handled on the hub (and the intention is to move this to the spoke as well).

The `cloud.Providerfactory` was modified to also return a boolean indicating whether or not a `Provider` exists.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stolostron/submariner-addon/pkg/cloud/gcp"
 	"github.com/stolostron/submariner-addon/pkg/cloud/rhos"
 	"github.com/stolostron/submariner-addon/pkg/constants"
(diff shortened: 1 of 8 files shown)


nyechiel:
@tpantelis this one seems to be ready for review. Should we try and get it into 2.6?

mkolesnik:
does this need cherry pick to 2.5?

skitt:
I think it does, we can discuss it on the cherry-pick

/cherrypick release-2.5

tpantelis:
> I think it does, we can discuss it on the cherry-pick
> 
> /cherrypick release-2.5

We don't need to cherry pick this - it builds on a prior PR so we'd need that one as well.

PR #454, opened 2022-07-25
Check OCP version for OVNK

vthapar:
OVNK requires OCP 4.11+ Check OCP version and set appropriate
status and error if unsupported version.

Note: We set status in Gateways node label coz it is used in
UI.

### DIFF
--- a/deploy/config/crds/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
+++ b/deploy/config/crds/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
@@ -206,6 +206,9 @@ spec:
                   infraId:
                     description: InfraId represents the infrastructure id of the managed cluster.
                     type: string
+                  networkType:
+                    description: NetworkType represents the network type (cni) of the managed cluster.
+                    type: string
                   platform:
                     description: Platform represents the cloud provider of the managed cluster.
                     type: string
@@ -214,6 +217,9 @@ spec:
                     type: string
                   vendor:
                     description: Vendor represents the kubernetes vendor of the managed cluster.
+                    type: string
+                  vendorVersion:
+                    description: VendorVersion represents the k8s vendor version of the managed cluster.
                     type: string
                 type: object
             type: object
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.11
+	github.com/coreos/go-semver v0.3.0
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/golang/mock v1.6.0
 	github.com/gophercloud/gophercloud v0.25.0
@@ -66,7 +67,6 @@ require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
-	github.com/coreos/go-semver v0.3.0 // indirect
 	github.com/coreos/go-systemd/v22 v22.3.3-0.20220203105225-a9a7ef127534 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dimchansky/utfbom v1.1.1 // indirect
--- a/pkg/spoke/submarineragent/config_controller.go
+++ b/pkg/spoke/submarineragent/config_controller.go
@@ -7,10 +7,12 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	operatorhelpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 	"github.com/pkg/errors"
+	"github.com/stolostron/submariner-addon/pkg/addon"
 	"github.com/stolostron/submariner-addon/pkg/apis/submarinerconfig"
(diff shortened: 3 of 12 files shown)


pkg/spoke/submarineragent/config_controller.go
```diff
@@ -553,6 +581,80 @@ func (c *submarinerConfigController) updateGatewayStatus(ctx context.Context, re
...
+			Message: fmt.Sprintf("OCP version is %s, submariner OVN requires 4.11.0+", ocpVersion),
+		}
+		updatedStatus, updated, err := addon.UpdateStatus(ctx, c.addOnClient, c.clusterName, addon.UpdateConditionFn(&ovnCondition))
+		if updated {
+			recorder.Eventf("ManagedClusterAddOnStatusUpdated", "Updated status conditions:  %#v",
+				updatedStatus.Conditions)
+		}
+
+		if err == nil {
+			err = errors.Errorf("OCP version is %s, submariner OVN requires 4.11.0+", ocpVersion)
```
skitt, line 649:
```suggestion
			err = errors.Errorf("OCP version is %s, Submariner OVN requires 4.11.0+", ocpVersion)
```

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/gophercloud/gophercloud v0.25.0
 	github.com/gophercloud/utils v0.0.0-20210909165623-d7085207ff6d
+	github.com/mcuadros/go-version v0.0.0-20190830083331-035f6764e8d2
```
skitt, line 11: This project is unmaintained, please use https://github.com/hashicorp/go-version instead.

skitt (changes requested):

deploy/config/crds/submarineraddon.open-cluster-management.io_submarinerconfigs.yaml
```diff
@@ -206,6 +206,9 @@ spec:
                   infraId:
                     description: InfraId represents the infrastructure id of the managed cluster.
                     type: string
+                  networkType:
+                    description: NetworkType represnts the network type (cni) of the managed cluster.
```
sridhargaddam, line 210:
```suggestion
                    description: NetworkType represents the network type (cni) of the managed cluster.
```

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -553,6 +581,80 @@ func (c *submarinerConfigController) updateGatewayStatus(ctx context.Context, re
...
+	}
+
+	if err != nil {
+		return false, err
+	}
+
+	networkType := config.Status.ManagedClusterInfo.NetworkType
+	ocpVersion := config.Status.ManagedClusterInfo.VendorVersion
+
+	if networkType == cni.OVNKubernetes && version.Compare(version.Normalize(ocpVersion), "4.11.0", "<") {
```
sridhargaddam, line 635: Please use a constant for the 4.11.0 similar to constants.ProductOCP

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -553,6 +581,80 @@ func (c *submarinerConfigController) updateGatewayStatus(ctx context.Context, re
...
+		ovnCondition := metav1.Condition{
+			Type:    "SubmarinerGatewayNodesLabeled",
```
sridhargaddam, line 637: Can we use the existing constant - "submarinerGatewayCondition"

sridhargaddam (changes requested):

pkg/spoke/submarineragent/config_controller.go
sridhargaddam: Is it because you are updating the ManagedClusterAddon and not SubmarinerConfig?

go.mod
vthapar: > This project is unmaintained, please use https://github.com/hashicorp/go-version instead.

Hashicorp is MPL 2.0 license. I used that first in subctl PR, but had to change due to failing license check.

skitt: So it is; how about https://github.com/blang/semver/v4?

skitt: … or https://github.com/coreos/go-semver/ which is already used elsewhere in Submariner.

pkg/spoke/submarineragent/config_controller.go
tpantelis: yeah - the `SubmarinerGatewayNodesLabeled` condition type checked in the UI. Ideally we shouldn't overload this condition. The `gatewaysStatusController` is responsible for maintaining this condition. It may be problematic also setting it here, ie if gateway nodes are still labeled then it could overwrite this setting. I think it would be safest (and perhaps cleaner) to put this code in the `gatewaysStatusController`.

tpantelis: Upon further discussion, using the `SubmarinerGatewayNodesLabeled` condition is the only direct way to inform the user so, unfortunately, we're limited. Also we want to elide cloud prepare so we can't do the check in the  `gatewaysStatusController`.

tpantelis: Let's use the constant:

```suggestion
			Type:    submarinerGatewayNodesLabeled,
```

go.mod
vthapar: Ran into this issue with semver:

```
OCP version is 4.11.0-rc.1, submariner OVN requires 4.11.0+
```
Should I use something like `4.11.0-rc` as min supported version instead of `4.11.0`?

sridhargaddam:
@vthapar unit tests seem to be failing.

skitt:
The `verify-deps` failure can be fixed by running `go mod tidy`.

skitt (approved):

PR #464, opened 2022-07-28
Fix missing secrets in SA for brokerInfo

### LINKED ISSUE #463: Submariner fails to deploy on OCP 4.11.0+
SubmarinerAddon Agent on hub cluster fails to create brokerInfo object with following error:

```
E0728 07:49:16.661477       1 base_controller.go:270] "SubmarinerAgentController" controller failed to sync "red/submariner", err: failed to create submariner brokerInfo of cluster red : ServiceAccount subm-broker/red does not have a secret of type token
```
Without brokerInfo object, submariner doesn't get deployed. This is issue with OCP 4.11.0+ because of change in how K8s 1.24+ creates Secrets for SAs.

vthapar:
Fixes #463

### DIFF
--- a/pkg/hub/submarinerbrokerinfo/brokerinfo.go
+++ b/pkg/hub/submarinerbrokerinfo/brokerinfo.go
@@ -18,6 +18,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

nyechiel:
@vthapar @skitt it looks like this was automatically synced into `release-2.6`, but should we also cherry pick to `release-2.5`?

vthapar:
Don't think so as this is mainly needed for 4.11 and 2.5 only supports up to 4.10.

nyechiel:
> Don't think so as this is mainly needed for 4.11 and 2.5 only supports up to 4.10.

2.5 should support 4.11 in a future 2.5.z.

PR #470, opened 2022-08-09
Update LB status in Submariner CR based on Config

sridhargaddam:
This PR updates the Loadbalancer status in the submariner CR
object based on the value in SubmarinerConfig object. This
will allow us to enable loadbalancer support by appropriately
setting the value in the SubmarinerConfig object.

### DIFF
--- a/pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml
+++ b/pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml
@@ -15,6 +15,7 @@ spec:
   ceIPSecPSK: {{ .IPSecPSK }}
   clusterCIDR: ""
   globalCIDR: "{{ .GlobalCIDR }}"
+  loadBalancerEnabled: "{{ .LoadBalancerEnabled }}"
   clusterID: {{ .ClusterName }}
   colorCodes: blue
   debug: false


sridhargaddam:
CC: @yboaron @vthapar @aswinsuryan

pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml
```diff
@@ -15,6 +15,7 @@ spec:
   ceIPSecPSK: {{ .IPSecPSK }}
   clusterCIDR: ""
   globalCIDR: "{{ .GlobalCIDR }}"
+  loadBalancerEnabled: "{{ .LoadBalancerEnabled }}"
```
yboaron, line 18: Not familiar with addon code, QQ: does this change support setting loadbalancer mode per cluster?

sridhargaddam:
CC: @skitt @tpantelis

pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml
tpantelis: yes this is the `Submariner` CR - this matches the exiting field in the [CRD](https://github.com/submariner-io/submariner-operator/blob/9bf9b0ba4eb131c2ad8c4c7b9005150cc46d662d/api/v1alpha1/submariner_types.go#L62).

yboaron: Ack, I wasn't sure how we get  {{ .LoadBalancerEnabled }} value, IIUC it's passesd at  [1] in brokerInfo parameter.
So, wanted to make sure we can set it per cluster 


[1]
https://github.com/stolostron/submariner-addon/blob/main/pkg/hub/submarineragent/controller.go#L514

tpantelis: `LoadBalancerEnabled` comes from the [SubmarinerConfig CR](https://github.com/stolostron/submariner-addon/blob/d1b28298d6dc3a53bc11469d7a84ea359f7ba20c/pkg/hub/submarinerbrokerinfo/brokerinfo.go#L186) which is per managed cluster.

nyechiel:
@sridhargaddam @yboaron note that without a cherry-pick to `release-2.6` this will only be available in 2.7. I think it's fine, unless QE wanted to give it a try before? /cc @MaxBab

PR #504, opened 2022-09-29
Bump to Submariner 0.13.1

skitt:


vendor/github.com/submariner-io/cloud-prepare/pkg/azure/ocpgwdeployer.go
```diff
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	submarinerGatewayGW      = "subgw"
+	submarinerGatewayGW      = "-subgw-"
```
sridhargaddam, line 42: Agree that this is part of the vendor folder, but it can create multiple GWs during upgrades. Do we handle that @aswinsuryan ?

hack/update-submariner.sh
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+for project in admiral cloud-prepare submariner submariner-operator; do
+    go get github.com/submariner-io/${project}@$1
+done
+
+sed -i "s/version: .*$/version: $1/" pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml
+sed -i "s/submver=.*$/submver=${1#v}/" scripts/deploy.sh
+
+go mod vendor
```
sridhargaddam, line 19: do we need "go mod tidy" as well?

vendor/github.com/submariner-io/cloud-prepare/pkg/azure/ocpgwdeployer.go
aswinsuryan: There is a check to if the active gateways match the required number of gateways, If yes cloud-prepare won't create a new g/w.

PR #522, opened 2022-10-17
Skip submariner operatorGroup

### LINKED ISSUE #475: Allow the operator to be overridden
Currently, the addon deploys the operator using an OLM catalog entry. This makes it difficult to deploy a custom operator for testing purposes.

We should add an entry in `SubmarinerConfig` to allow the operator image to be specified. When that’s populated, the addon would _not_ deploy using the catalog, but deploy the operator directly instead. It will also need to take care of all the surrounding infrastructure (CRDs, RBAC etc.), either by relying on the relevant code from `subctl`, or by building a manifest from the YAML files already present in the addon and deploying that (through a `ManifestWork`).

vthapar:
Adds an option to skip installation of `submariner-operator` OperatorGroup if annotation `skipOperatorGroup` is set on the SubmarinerConfig. This will allow dev users to override submariner-operator Image or any other component installed.

How To Use custom operator image:

1. Install submariner-addon on managedcluster in the normal way.
2. Once installed, add annotation `skipOperatorGroup` to the SubmarinerConfig for the managedcluster where you wish to use custom operator Image or settings.
3. Modify `submariner-operator` Deployment to use whichever Image you want.

Note: This feature is for dev use only.

Fixes #475

### DIFF
--- a/pkg/hub/submarineragent/controller.go
+++ b/pkg/hub/submarineragent/controller.go
@@ -78,8 +78,12 @@ var sccFiles = []string{
 	"manifests/rbac/submariner-agent-scc.yaml",
 }
 
-var operatorFiles = []string{
+var operatorAllFiles = []string{
 	"manifests/operator/submariner-operator-group.yaml",
+	"manifests/operator/submariner-operator-subscription.yaml",
+}
+
+var operatorSkipFiles = []string{
 	"manifests/operator/submariner-operator-subscription.yaml",
 }
 
@@ -498,15 +502,19 @@ func (c *submarinerAgentController) deploySubmarinerAgent(
 		return fmt.Errorf("failed to create submariner brokerInfo of cluster %v : %w", managedCluster.Name, err)
 	}
 
+	skipOperatorGroup := false
+
 	if submarinerConfig != nil {
 		err := c.updateSubmarinerConfigStatus(ctx, submarinerConfig, managedCluster.Name)
 		if err != nil {
 			return err
 		}
+
+		_, skipOperatorGroup = submarinerConfig.GetAnnotations()["skipOperatorGroup"]
 	}
 
 	// Apply submariner operator manifest work
-	operatorManifestWork, err := newOperatorManifestWork(managedCluster, brokerInfo)
+	operatorManifestWork, err := newOperatorManifestWork(managedCluster, brokerInfo, skipOperatorGroup)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/hub/submarineragent/controller.go
```diff
@@ -657,13 +659,17 @@ func newSubmarinerManifestWork(managedCluster *clusterv1.ManagedCluster, config
...
-func newOperatorManifestWork(managedCluster *clusterv1.ManagedCluster, config interface{}) (*workv1.ManifestWork, error) {
+func newOperatorManifestWork(managedCluster *clusterv1.ManagedCluster, config interface{}, skipOperatorGroup bool) (*workv1.ManifestWork, error) {
 	files := []string{agentRBACFile}
 	if getClusterProduct(managedCluster) == constants.ProductOCP {
 		files = append(files, sccFiles...)
 	}
 
-	files = append(files, operatorFiles...)
+	if skipOperatorGroup {
+		files = append(files, operatorFiles[1:]...)
```
skitt, line 669: This is rather fragile, if we ever update or re-order `operatorFiles` it’s likely that we won’t remember to change this. Please split up `operatorFiles` and handle the two entries separately — something like

```
const operatorGroupFile = "manifests/operator/submariner-operator-group.yaml"
const operatorSubscriptionFile = "manifests/operator/submariner-operator-subscription.yaml"

// ...

if !skipOperatorGroup {
    files = append(files, operatorGroupFile)
}

files = append(files, operatorSubscriptionFile)
```

skitt (changes requested):

pkg/hub/submarineragent/controller.go
vthapar: Funnily enough, that is the way I had done initially. Changed ot this to reduce consts/vars. I was considering adding a note to `operatorFiles` declaration to always keep `submariner-operator-group.yaml` at `[0]`.

pkg/hub/submarineragent/controller.go
```diff
@@ -77,9 +78,9 @@ var sccFiles = []string{
 	"manifests/rbac/submariner-agent-scc.yaml",
 }
 
-var operatorFiles = []string{
-	"manifests/operator/submariner-operator-group.yaml",
-	"manifests/operator/submariner-operator-subscription.yaml",
+var operatorFiles = map[string]string{
+	operatorGroup:       "manifests/operator/submariner-operator-group.yaml",
+	"subscriptionGroup": "manifests/operator/submariner-operator-subscription.yaml",
```
skitt, line 83: I’m not convinced by this approach — it requires both a specific data structure _and_ supporting code.

If you want to keep the idea of lists of files, how about

```
var operatorAllFiles = [] string {
    "manifests/operator/submariner-operator-group.yaml",
    "manifests/operator/submariner-operator-subscription.yaml",
}

var operatorSkipFiles = []string {
    "manifests/operator/submariner-operator-subscription.yaml",
}

...

if skipOperatorGroup {
    files = append(files, operatorSkipFiles)
} else {
    files = append(files, operatorAllFiles)
}
```

Or we could make the whole thing fully generic: make a “skip” annotation which is a list of files to skip, and when populating the lists of files, check that each file isn’t in the annotation.

skitt (changes requested):

pkg/hub/submarineragent/controller.go
vthapar: I like the two vars with different set of files idea. I want to add skip annotation with files coz it adds more work for devs and needs them to be aware of what are the file paths to use. We can go that way if we tomorrow have more files and want to let devs pick and choose.

Issue #529, opened 2022-10-24
ARO: SubmarinerConfig status set to false with reason 'SubmarinerClusterEnvPreparationFailed'


yboaron:
I imported ARO cluster to ACM using kubeconfig file and next step applied submariner-addon .
Deployment is stuck with [1] error in SubmarinerConfig status.
 

[1]
status:
  conditions:
  - lastTransitionTime: "2022-10-23T15:16:43Z"
    message: SubmarinerConfig was applied
    reason: SubmarinerConfigApplied
    status: "True"
    type: SubmarinerConfigApplied
  - lastTransitionTime: "2022-10-23T15:16:49Z"
    message: 'Failed to prepare submariner cluster environment: unsupported vendor
      "ARO" of cluster "aro-cluster"'
    reason: SubmarinerClusterEnvPreparationFailed
    status: "False"
    type: SubmarinerClusterEnvironmentPrepared
  managedClusterInfo:
    clusterName: aro-cluster
    infraId: yb-cluster-4z7xb
    platform: Azure
    region: eastus
    vendor: ARO
    vendorVersion: 4.10.20

PR #530, opened 2022-10-24
Add support for ARO

yboaron:
For ARO managed cluster:
 - OCP scc files should be applied
 - cloud prepare should be skipped

Fixes: https://github.com/stolostron/submariner-addon/issues/529

Signed-off-by: Yossi Boaron <yboaron@redhat.com>


yboaron:
Checked it from ACM - it looks OK.

PR #533, opened 2022-10-25
Bump to K8s version 25.0

tpantelis:
Signed-off-by: Tom Pantelis <tompantelis@gmail.com>


skitt:
/hold

This now passes with the updated `golangci-lint`, @tpantelis can you clean the PR up? The first two commits should be squashed, and presumably the last one dropped.

tpantelis:
> This now passes with the updated `golangci-lint`, @tpantelis can you clean the PR up? The first two commits should be squashed, and presumably the last one dropped.

Done.

PR #585, opened 2022-11-15
Restart addon pod on managed cluster after restore

### LINKED ISSUE #584: [2.7] Restart submariner-addon pod on managed cluster after restore
This is required for https://github.com/stolostron/backlog/issues/22005

When hub is backed up and restored on backup cluster, the submariner-adodn pods on managed clusters continue to use old hub. To make them use new hub, need restart submariner-addon pods.

vthapar:
After a backup+restore, submariner-addon pods on managed cluster are still trying to reach old hub. The pods need to be restarted to pick up new hub config.

Add an env variable to submariner-addon deployment with hub's API server. When new hub is restored, it will result in forced restart of the pods on managed cluster.

Fixes #584

### DIFF
--- a/pkg/hub/manager.go
+++ b/pkg/hub/manager.go
@@ -186,7 +186,14 @@ func (o *AddOnOptions) RunControllerManager(ctx context.Context, controllerConte
 		return err
 	}
 
-	err = mgr.AddAgent(submarineraddonagent.NewAddOnAgent(kubeClient, controllerContext.EventRecorder, o.AgentImage))
+	localCluster, err := clusterClient.ClusterV1().ManagedClusters().Get(context.TODO(), "local-cluster", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	host := localCluster.Spec.ManagedClusterClientConfigs[0].URL
+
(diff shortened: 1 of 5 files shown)


skitt:
I take it the idea here is that a change in host hub will result in a change in the pod descriptor, and thus a restart of the pod, is that right?

vthapar:
> I take it the idea here is that a change in host hub will result in a change in the pod descriptor, and thus a restart of the pod, is that right?

Yes. The value in env variable will change and this change in deployment will result in pod restart. I found this to be cleanest and simplest solution. Some other options I tried or considered:

1. Check for informer/listers getting errors and recreate the hubclient. This approach is used in another addon [I think search addon] but doesn't work for us coz we use multiple controllers taking hub kubeclient as arguent with no access to hub config.

2. Check for errors in informer/listers and restart the pod. Better than 1 but tricky in implementation. Even a panic won't force pod restart coz all controllers run in separate threads. Also, the error we get is Timeout which can also happen if network is slow and api server times out. False negatives can lead to unnecessary restarts.

This works best coz this change will be trigerred by addon code running on hub and it will only change value in case of backup/restore, avoiding any false negatives.

PR #594, opened 2022-11-18
Fix race between addon pods and local-cluster

vthapar:
On bring up, addon pod on hub looks for local-cluster to fetch API URL. Since addon pod comes up during cluster creation, it is possible that `local-cluster` isn't available yet. This can result in hubHost to be empty and breaks backup/restore.


pkg/hub/submarineraddonagent/agent.go
```diff
@@ -231,3 +242,22 @@ func (a *addOnAgent) permissionConfig(cluster *clusterv1.ManagedCluster, _ *addo
 
 	return operatorhelpers.NewMultiLineAggregate(errs)
 }
+
+func (a *addOnAgent) setHubHostIfEmpty() error {
+	if a.hubHost == "" {
```
tpantelis, line 247: Why would this be called multiple times to warrant this check? Also how does moving the code here fix the race condition? Can you add a comment explaining the rationale?

vthapar: Earlier this code was sitting in `manager.go` and ran only the first time pod came up. This code is now in `agent.go#Manifest()` and gets called when submariner is installed. Also, this bit of code is repeatedly called by Syncer to check if anything changed. So, even if local cluster was missing at bring up, it will be available by the time submariner is installed. 

In case of kind setup, there will never be local-cluster, so this will always be empty. But in case of actual deployments in OCP, this will only be empty first time. This check helps skip the repeated `Get()` calls if it is already set while at same time lets us query if it was for some reason empty.

One scenario where local cluster might be missing with submariner present is in case of hub restart [maybe coming out of hibernation] but since this code gets called repeatedly by syncer, it will correct in subsequent iteration.

tpantelis: Thanks for the explanation. Can you add this as a comment in he code?

PR #606, opened 2022-11-28
Fix Spoke re-deploy

mkolesnik:
Fix Spoke code to handle changes in the `SubmarinerConfig` instead of ignoring them if the cluster has been deployed.

Signed-off-by: Mike Kolesnik <mkolesni@redhat.com>

### DIFF
--- a/pkg/spoke/submarineragent/config_controller.go
+++ b/pkg/spoke/submarineragent/config_controller.go
@@ -3,6 +3,7 @@ package submarineragent
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/spoke/submarineragent/config_controller.go
```diff
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	corev1lister "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/klog"
```
sridhargaddam, line 38: We use "k8s.io/klog/v2" at other places in the code. Can we use the same even here instead of adding new dependency on klog?

pkg/spoke/submarineragent/config_controller.go
```diff
@@ -186,13 +191,29 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
 		return updateErr
 	}
 
+	return c.syncConfig(ctx, recorder, config)
```
sridhargaddam, line 194: The commit message says that its for GCP Clusters, but the code seems to be applicable for the other clusters as well. Can you please clarify/update the commit message?

pkg/spoke/submarineragent/config_controller.go
mkolesnik: Sure, it was auto-imported, I updated it to v2

pkg/spoke/submarineragent/config_controller.go
mkolesnik: Sure I updated the commit message

skitt:
@mkolesnik could you take care of the manual cherry-pick on `release-2.5`?

mkolesnik:
Manually cherry picked as #617

PR #607, opened 2022-11-28
Use correct version of submariner images in kind

vthapar:
When deploying addon in kind, it ignores the submariner repositrory and version. This causes a discrepancy where operator is using older version [0.11.0] while all agents are on latest version configured. Also, for route agent and lighthouse images, overrides don't work correctly. This is because component name is incorrect.

This fix requires following changes:
 1. Wait for all submariner manifestworks to be created and available.
 2. Set `skipOperatorGroup` annotation on submarinerconfigs
 3. Patch submariner-operator deployments to use repo and version as configured.
 4. Use correct component names for all lighthouse and routeagent image overrides.

### DIFF
--- a/scripts/demo.sh
+++ b/scripts/demo.sh
@@ -3,6 +3,7 @@
 set -o nounset
 
 source scripts/clusters.sh
+source scripts/vars.sh
 
 hub="${clusters[0]}"
 managedcluster1="${clusters[0]}"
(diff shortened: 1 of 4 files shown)


scripts/demo.sh
```diff
@@ -14,6 +15,34 @@ kubeconfigs_dir="${work_dir}/kubeconfigs"
...
+    local cluster=$1
+    echo "Waiting for submariner manifestworks to be ready for ${cluster}"
+    kubectl wait --for=condition=available manifestworks submariner-operator -n ${cluster} --timeout=60s
+    kubectl wait --for=condition=available manifestworks submariner-resource -n ${cluster} --timeout=60s
+
+    echo "Waiting to patch submariner-operator deployment on ${cluster}"
+    kubectl annotate submarinerconfig submariner -n ${cluster} skipOperatorGroup=""
+
+    submrepo="quay.io/submariner"
+    submver=0.14.0-rc2
```
skitt, line 28: Why are these overridden here?

scripts/vars.sh
```diff
@@ -0,0 +1,3 @@
+k8s_version="v1.23.4"
+submrepo="quay.io/submariner"
+submver="0.14.0-rc2"
```
skitt, line 3: Let’s bump to 0.14.0 while we’re at it.

skitt (changes requested):

scripts/demo.sh
sridhargaddam, line 28: Why not 0.14.0?

vthapar: Ha, forgot to update. Was testing scripts before 0.14.0 as available :)

sridhargaddam: Also this is redundant as we are declaring and sourcing them from scripts/vars.sh

vthapar: Was testing couple things for gateway issue, missed cleanup.

skitt (approved):

PR #610, opened 2022-11-29
Don't open vxlan port for OVN

### LINKED ISSUE #388: Add OVN support
When using OVNKubernetes as CNI we don't need to open all the vxlan ports in the cloud. So only open the vxlan ports required based on which CNI is used during cloud prepare step.

1. Detect if OVNKubernetes is the CNI
2. If yes, open the minimal set of ports.

vthapar:
Submariner doesn't use VxLan port [4800] when CNI is OVNKubernetes. So no need to open this port in cloud.

Fixes #388 

### DIFF
--- a/pkg/cloud/aws/aws.go
+++ b/pkg/cloud/aws/aws.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/stolostron/submariner-addon/pkg/cloud/provider"
 	"github.com/stolostron/submariner-addon/pkg/cloud/reporter"
@@ -11,6 +12,7 @@ import (
 	cpaws "github.com/submariner-io/cloud-prepare/pkg/aws"
 	cpclient "github.com/submariner-io/cloud-prepare/pkg/aws/client"
 	"github.com/submariner-io/cloud-prepare/pkg/ocp"
+	"github.com/submariner-io/submariner/pkg/cni"
 )
 
 const (
@@ -26,6 +28,7 @@ type awsProvider struct {
 	nattPort          int64
 	nattDiscoveryPort int64
 	instanceType      string
+	cniType           string
 	gateways          int
 	cloudPrepare      cpapi.Cloud
 	gatewayDeployer   cpapi.GatewayDeployer
@@ -78,6 +81,7 @@ func NewProvider(info *provider.Info) (*awsProvider, error) {
 		reporter:          reporter.NewEventRecorderWrapper("AWSCloudProvider", info.EventRecorder),
 		nattPort:          int64(info.IPSecNATTPort),
 		nattDiscoveryPort: int64(info.NATTDiscoveryPort),
+		cniType:           info.NetworkType,
 		instanceType:      instanceType,
 		gateways:          info.Gateways,
 		cloudPrepare:      cloudPrepare,
@@ -101,10 +105,13 @@ func (a *awsProvider) PrepareSubmarinerClusterEnv() error {
 	}, a.reporter); err != nil {
 		return err
 	}
-	if err := a.cloudPrepare.OpenPorts([]cpapi.PortSpec{
-		{Port: constants.SubmarinerRoutePort, Protocol: "udp"},
-	}, a.reporter); err != nil {
-		return err
+
+	if !strings.EqualFold(a.cniType, cni.OVNKubernetes) {
+		if err := a.cloudPrepare.OpenPorts([]cpapi.PortSpec{
+			{Port: constants.SubmarinerRoutePort, Protocol: "udp"},
+		}, a.reporter); err != nil {
+			return err
+		}
 	}
 
 	a.reporter.Success("The Submariner cluster environment has been set up on AWS")
--- a/pkg/cloud/azure/azure.go
+++ b/pkg/cloud/azure/azure.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"strconv"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/pkg/errors"
@@ -16,6 +16,7 @@ import (
 	"github.com/submariner-io/cloud-prepare/pkg/azure"
 	"github.com/submariner-io/cloud-prepare/pkg/k8s"
 	"github.com/submariner-io/cloud-prepare/pkg/ocp"
+	"github.com/submariner-io/submariner/pkg/cni"
 	corev1 "k8s.io/api/core/v1"
 )
 
@@ -27,7 +28,7 @@ const (
 type azureProvider struct {
 	infraID           string
 	nattPort          uint16
-	routePort         string
+	cniType           string
 	cloudPrepare      api.Cloud
 	reporter          submreporter.Interface
 	gwDeployer        api.GatewayDeployer
@@ -84,7 +85,7 @@ func NewProvider(info *provider.Info) (*azureProvider, error) {
 	return &azureProvider{
 		infraID:           info.InfraID,
 		nattPort:          uint16(info.IPSecNATTPort),
-		routePort:         strconv.Itoa(constants.SubmarinerRoutePort),
+		cniType:           info.NetworkType,
 		cloudPrepare:      cloudPrepare,
 		gwDeployer:        gwDeployer,
 		reporter:          reporter.NewEventRecorderWrapper("AzureCloudProvider", info.EventRecorder),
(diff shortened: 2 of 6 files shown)


pkg/cloud/aws/aws.go
```diff
@@ -119,12 +122,15 @@ func (a *awsProvider) PrepareSubmarinerClusterEnv() error {
...
+
+	if a.cniType != cni.OVNKubernetes {
+		if err := a.cloudPrepare.PrepareForSubmariner(cpapi.PrepareForSubmarinerInput{
+			InternalPorts: []cpapi.PortSpec{
+				{Port: constants.SubmarinerRoutePort, Protocol: "udp"},
+			},
+		}, a.reporter); err != nil {
+			return err
+		}
 	}
```
skitt, line 134: Is the difference in behaviour here intentional? For the other providers, `PrepareForSubmariner` is called anyway, here it isn’t called at all.

skitt (changes requested):

pkg/cloud/aws/aws.go
vthapar: @skitt I wasn't sure which of the two is preferred here. So decided to write code this way so it comes up in review and get a better answer. All `PrepareForSubmariner` does today is add SG rules. So, do we need to call if the list is empty, or we still call it in case we add more stuff to it in `cloud` later down the line.

@sridhargaddam Thoughts?

nyechiel:
@vthapar can you please address the review comments? I want to ensure QE gets enough time to test it for 2.7. Thanks!

vthapar:
Found an issue in testing. Fix is ready, testing it out and will push once confirmed.

aswinsuryan:
> Here is the core of problem and why this fix in current form doesn't work. This is the current sequence:
> 
> 1. SubmarinerAddon agent on hub cluster reads `ClusterClaims` on `ManagedCluster` object to get information like platform, Vendor, Version etc. This information is added to `SubmarinerConfig.Status.ManagedClusterInfo`. After it, add on does cloud prepare.
> 2. SubmarinerAddon agent on spoke checks if ManagedCluserInfo is present or not. If not, it just returns. If present it fetches `NetworkType` from `Networks` object which only exists on spoke clusters. This is what tells us which CNI it is. It also adds NetworkType to `SubmarinerConfig.Status.ManagedClusterInfo`.
> 3. The fix requires NetworkType information to be available in `SubmarinerConfig` before we do cloud prepare. Note that NetworkType is not available on `ManagedCluster` object, even as `ClusterClaims`.
> 4. So, we're in a kind of catch22 for the fix. Hub requires information on NetworkType before it does cloud prepare and adds `ManagedClusterInfo` to `SubmarinerConfig`. Spoke requires `ManagedClusterInfo` to be present before it, amongst other things, adds `NetworkType` to `ManagedClusterInfo`.
> 
> This is the possible fix am going for:
> 
> 1. On the Hub, if `NetworkType` is missing from `SubmarinerConfig.Status`, return error so we can retry.
> 2. On Spoke if ManagedClusterInfo is missing, set it with `NetworkType` set and return. Do not do any other operations.
> 3. Hub will eventually find `NetworkType` on `SubmarinerConfig` and add platform, product, version etc. to it and proceed with cloud prepare based on network type.
> 4. Spoke will eventuall find Platform etc. available on `SubmarinerConfig` and proceed with its operations.
> 
> I am testing out this fix and once I confirm it to be working, will push it.
> 
> Ideal fix would be to have `NetowrkType` available on ManagedCluster object itself, but we don't own it. Maybe raise an enhancement request for it?

To add it, the cloud prepare happens in hub only for AWS cluster, for all others it happens in spoke. So we may have to test the fix in both.

tpantelis:
> It has been decided to simplify this fix so that it works for non AWS. It will start working for AWS once #218 is fixed.

@nyechiel ☝️ We have a use case where AWS being on the hub is actually problematic functionally and not just for code maintenance. I can work on the code changes rest of this week and next which should be enough time (after that I'm away on PTO).  But if not, @aswinsuryan et al can finish it.

sridhargaddam:
> Had a discussion on this in the team meeting and here are results:
> 
> As @aswinsuryan pointed out, this is an issue only for AWS which runs cloud prepare on hub. Since this change adds a significant delay to cloud prepare [2-3 minutes] it would be better to wait for #218 to be fixed. Then this fix would be much simpler.
> 
> It has been decided to simplify this fix so that it works for non AWS. It will start working for AWS once #218 is fixed.

I guess you wanted to tag me, as @aswinsuryan was not in the call ;-)

nyechiel:
> > It has been decided to simplify this fix so that it works for non AWS. It will start working for AWS once #218 is fixed.
> 
> @nyechiel point_up We have a use case where AWS being on the hub is actually problematic functionally and not just for code maintenance. I can work on the code changes rest of this week and next which should be enough time (after that I'm away on PTO). But if not, @aswinsuryan et al can finish it.

I will defer to you folks, but note that: 

- ACM is already in feature freeze for 2.7
- OVN is the default CNI in OpenShift now
- AWS is probably the most commonly used platform

My understanding is that https://github.com/stolostron/submariner-addon/pull/610 is not _required_ for OVN support. It's more of a security optimization to ensure we don't open unused ports.

So while I am supportive of the idea to address all of this for 2.7 in the next couple of weeks, let's also by mindful of the risk here, and feel empowered to push this to 2.8 at any point if we feel like this is the right thing to do.

@skitt @sridhargaddam FYI

tpantelis:
> My understanding is that #610 is not _required_ for OVN support. It's more of a security optimization to ensure we don't open unused ports.

If that's case then let's defer this PR as well.

nyechiel:
> > ACM is already in feature freeze for 2.7
> 
> #218 is really tech debt and not a feature.
> 

Indeed. Technically we can get it in before Dec 31, but my point is that we are still very late in the cycle, and just got into a point where we have stable builds for QE...

> > My understanding is that #610 is not _required_ for OVN support. It's more of a security optimization to ensure we don't open unused ports.
> 
> If that's case then let's defer this PR as well.

tpantelis:
> > #218 is really tech debt and not a feature.
> 
> Indeed. Technically we can get it in before Dec 31, but my point is that we are still very late in the cycle, and just got into a point where we have stable builds for QE...

Understood. In looking at the AWS code, it's basically similar to the other platforms in that it creates a machine set - the difference is that AWS deploys it using a manifest work b/c it's on the hub whereas the others use a direct cluster client. So it looks like the changes are pretty straightforward.  Everything carries some risk - in this case either it works or it doesn't so, as long as we can verify it before merging, I think the risk is minimal.  That said, I think we should only get that in for 2.7 if we want to get this PR in.

nyechiel:
> > > #218 is really tech debt and not a feature.
> > 
> > 
> > Indeed. Technically we can get it in before Dec 31, but my point is that we are still very late in the cycle, and just got into a point where we have stable builds for QE...
> 
> Understood. In looking at the AWS code, it's basically similar to the other platforms in that it creates a machine set - the difference is that AWS deploys it using a manifest work b/c it's on the hub whereas the others use a direct cluster client. So it looks like the changes are pretty straightforward. Everything carries some risk - in this case either it works or it doesn't so, as long as we can verify it before merging, I think the risk is minimal. That said, I think we should only get that in for 2.7 if we want to get this PR in.

@vthapar @tpantelis so what's the conclusion here? Do we want to try and target both https://github.com/stolostron/submariner-addon/pull/610 and https://github.com/stolostron/submariner-addon/issues/218 for 2.7?

tpantelis:
> @vthapar @tpantelis so what's the conclusion here? Do we want to try and target both #610 and #218 for 2.7?

As mentioned, #610 depends on #218 so is #610 needed for 2.7? @sridhargaddam

sridhargaddam:
> As mentioned, #610 depends on #218 so is #610 needed for 2.7? @sridhargaddam

I understand the concern from @nyechiel. We managed to get stable builds since last few days and with holidays coming up, we may not have enough time to validate all the scenarios on AWS (including ROSA/ARO etc). Also, as Nir pointed out #610 is more like a security optimisation and considering that the port is an internal port and not something that we open for outside world, I'd say that we can defer both the PRs.

nyechiel:
> > As mentioned, #610 depends on #218 so is #610 needed for 2.7? @sridhargaddam
> 
> I understand the concern from @nyechiel. We managed to get stable builds since last few days and with holidays coming up, we may not have enough time to validate all the scenarios on AWS (including ROSA/ARO etc). Also, as Nir pointed out #610 is more like a security optimisation and considering that the port is an internal port and not something that we open for outside world, I'd say that we can defer both the PRs.

Since we seem to have some capacity now and the topic is fresh in your mind, may I suggest that you (Tom and Vishal) start working on both PRs now but for 2.8? AFAIK, nothing should stop us from creating a `release-2.8` branch now (we should have full control on this for the `submariner-addon` repo).

tpantelis:
> Since we seem to have some capacity now and the topic is fresh in your mind, may I suggest that you (Tom and Vishal) start working on both PRs now but for 2.8? AFAIK, nothing should stop us from creating a `release-2.8` branch now (we should have full control on this for the `submariner-addon` repo).

👍 I was going to work on https://github.com/stolostron/submariner-addon/issues/218 next week regardless.

vthapar:
@nyechiel #610 will work for non AWS even without #218 I agree with @sridhargaddam that it is not that high priority, but in case someone raises an issue and we need it in 2.7.1 or a z-stream later, we can still bring it in for non AWS platforms without backporting #218

nyechiel:
> @nyechiel #610 will work for non AWS even without #218 I agree with @sridhargaddam that it is not that high priority, but in case someone raises an issue and we need it in 2.7.1 or a z-stream later, we can still bring it in for non AWS platforms without backporting #218

Ok, thanks for the info. I just discussed with @skitt and he is going to create the `release-2.8` branch in the addon. Let's get all those PRs worked on there.

vthapar:
Converted to draft to avoid merge and reviews. Will revisit this sfter #218 is merged so can test e2e and make sure this works fine.

vthapar:
Opened for review after basic testing on AWS and GCP with OVNKubernetes. Will continue doing more testing while PR is under review. Could use help with testing on RHOS and Azure.

PR #631, opened 2022-12-09
Drop the version tag from the Submariner CR

skitt:
The bundle used for Submariner contains the relevant image hashes, so we shouldn't need to override the version with a tag any more. Specifying the tag means that OpenShift's disconnected setup features can't be used; removing the tag should enable them.

### DIFF
--- a/pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml
+++ b/pkg/hub/submarineragent/manifests/operator/submariner.io-submariners-cr.yaml
@@ -53,4 +53,3 @@ spec:
     {{- end}}
 {{- end}}
   repository: registry.redhat.io/rhacm2
-  version: v0.14.0


sridhargaddam:
> The bundle used for Submariner contains the relevant image hashes, so we shouldn't need to override the version with a tag any more. Specifying the tag means that OpenShift's disconnected setup features can't be used; removing the tag should enable them.

@skitt is this the fix for supporting image downloads using the digest instead of tags?

skitt:
> > The bundle used for Submariner contains the relevant image hashes, so we shouldn't need to override the version with a tag any more. Specifying the tag means that OpenShift's disconnected setup features can't be used; removing the tag should enable them.
> 
> @skitt is this the fix for supporting image downloads using the digest instead of tags?

Hopefully, yes!

nyechiel:
> > > The bundle used for Submariner contains the relevant image hashes, so we shouldn't need to override the version with a tag any more. Specifying the tag means that OpenShift's disconnected setup features can't be used; removing the tag should enable them.
> > 
> > 
> > @skitt is this the fix for supporting image downloads using the digest instead of tags?
> 
> Hopefully, yes!

For reference, here is the issue: https://github.com/stolostron/backlog/issues/23726

MaxBab:
Great.
This should help us in testing ROSA as well.

PR #705, opened 2023-02-15
Fix the default GW instance type for Submariner GW

sridhargaddam:
This PR updates the default Submariner Gateway instance type for Azure/RHOS clusters to the one that is actually deployed.


sridhargaddam:
CC @aswinsuryan @tpantelis @skitt

aswinsuryan (approved):

PR #768, opened 2023-03-15
Bump to Submariner 0.15.0-m4

maayanf24:


go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	k8s.io/code-generator v0.26.2
 	k8s.io/component-base v0.26.2
 	k8s.io/klog/v2 v2.90.1
-	open-cluster-management.io/addon-framework v0.6.1
+	open-cluster-management.io/addon-framework v0.1.0
```
skitt, line 35: I’m not sure why this gets downgraded, but it shouldn’t be.

skitt (changes requested):

skitt:
@maayanf24 to fix this, you need to re-upgrade the addon-framework:

```
go get open-cluster-management.io/addon-framework
go mod tidy
```

This will restore the missing `go.sum` entries that were removed in the downgrade.

skitt (approved):

PR #806, opened 2023-04-03
Bump submariner to 0.15.0-rc0

vthapar:


go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	k8s.io/code-generator v0.26.3
 	k8s.io/component-base v0.26.3
 	k8s.io/klog/v2 v2.90.1
-	open-cluster-management.io/addon-framework v0.6.1
+	open-cluster-management.io/addon-framework v0.1.0
```
vthapar, line 35: @skitt Any idea why this change? I used the `update-submariner.sh` script.

skitt: No, I haven’t figured out why `go get` downgrades that dependency... You need to manually fix it.

skitt:
This is failing because of the status update changes in Admiral. We can fix it either by populating the default scheme, or waiting for the upstream fix to be backported to the 0.15 branch and updating to that.

tpantelis:
> This is failing because of the status update changes in Admiral. We can fix it either by populating the default scheme, or waiting for the upstream fix to be backported to the 0.15 branch and updating to that.

Just need another approval on https://github.com/submariner-io/admiral/pull/593.  @vthapar @sridhargaddam @aswinsuryan

PR #830, opened 2023-04-17
Switch to a Go 1.20 builder

skitt:


skitt:
I’m not sure this is sufficient to get Go 1.20 code building, I think we also need to update our `openshift/release` configuration.

skitt:
The `o/r` PR is https://github.com/openshift/release/pull/38411.

tpantelis (approved):

PR #871, opened 2023-05-09
Add backup label to globalnet configmap

vthapar:
When creating globalnet configmap, add backup label to it so that it is backed up to new hub. Without this label, addon agent will reallocate global cidrs to managedclusters on restore. If clusters get new CIDRs allocated, it will break connectivity.

### DIFF
--- a/pkg/hub/submarineragent/controller.go
+++ b/pkg/hub/submarineragent/controller.go
@@ -63,7 +63,8 @@ const (
 	submarinerCRFile              = "manifests/operator/submariner.io-submariners-cr.yaml"
 	BrokerCfgApplied              = "SubmarinerBrokerConfigApplied"
 	BrokerObjectName              = "submariner-broker"
-	BackupLabel                   = "cluster.open-cluster-management.io/backup"
+	BackupLabelKey                = "cluster.open-cluster-management.io/backup"
+	BackupLabelValue              = "submariner"
 	addonDeploymentConfigResource = "addondeploymentconfigs"
 	addonDeploymentConfigGroup    = "addon.open-cluster-management.io"
 )
@@ -663,7 +664,8 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(ctx context.Con
 	}
 
 	if gnCmErr == nil {
-		return nil
+		// This should handle upgrade from a version that didn't add the label
+		return c.updateBackupLabelOnGnConfigMap(ctx, brokerNamespace)
 	}
 
 	// globalnetConfig is missing in the broker-namespace, try creating it from submariner-broker object.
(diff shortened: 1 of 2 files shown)


pkg/hub/submarineragent/controller.go
```diff
@@ -692,9 +692,33 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(ctx context.Con
 		return errors.Wrapf(err, "error creating globalnet configmap on Broker")
 	}
 
+	// Ideally we should add option to have custom labels when creating configmap
```
skitt, line 695: Please open an issue on the operator to do this.

pkg/hub/submarineragent/controller.go
```diff
@@ -692,9 +692,33 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(ctx context.Con
...
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+
+	if gnConfigMap != nil {
+		gnConfigMap.ObjectMeta.Labels[BackupLabel] = "submariner"
+		err = c.controllerClient.Update(ctx, gnConfigMap)
+	}
+
+	return err
```
skitt, line 719: Shouldn’t this be in a retry loop?

skitt (changes requested):

pkg/hub/submarineragent/controller.go
vthapar: We are the only writer of this resource and caller will end up retrying so didn't do that. Still, good idea to retry here than repeat the entire sync loop.

pkg/hub/submarineragent/controller.go
vthapar: https://github.com/submariner-io/submariner-operator/issues/2636

pkg/hub/submarineragent/controller.go
sridhargaddam: If there is an error while adding the label, the sync logic will not help. When the code finds that the configMap already exists, it will return in Line 666

tpantelis (changes requested):

pkg/hub/submarineragent/controller.go
```diff
@@ -692,6 +697,40 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(ctx context.Con
...
+		}
+
+		return err
+	})
+
+	if retryErr != nil {
+		return errors.Wrapf(retryErr, "error adding backup label to globalnet configmap in %q", namespace)
+	}
+
 	return nil
```
skitt, line 734:
```suggestion
	return errors.Wrapf(retry.RetryOnConflict(retry.DefaultRetry, func() error {
		gnConfigMap, err := globalnet.GetConfigMap(ctx, c.controllerClient, namespace)
		if apierrors.IsNotFound(err) {
			return nil
		}

		if err != nil {
			return err
		}

		labels := gnConfigMap.GetLabels()
		if labels == nil {
			labels = make(map[string]string)
		}

		if _, ok := labels[BackupLabel]; !ok {
			labels[BackupLabel] = "submariner"
			gnConfigMap.SetLabels(labels)
			err = c.controllerClient.Update(ctx, gnConfigMap)
			if err == nil {
				logger.Infof("Successfully added backup label to submariner-broker in %q", namespace)
			}
		}

		return err
	}), "error adding backup label to globalnet configmap in %q", namespace)
```

skitt (changes requested):

tpantelis (approved):

pkg/hub/submarineragent/controller.go
```diff
@@ -692,7 +693,37 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(ctx context.Con
 		return errors.Wrapf(err, "error creating globalnet configmap on Broker")
 	}
 
-	return nil
+	// Ideally we should add option to have custom labels when creating configmap
+	return c.updateBackupLabelOnGnConfigMap(ctx, brokerNamespace)
```
tpantelis, line 697: We don't actually need this. We can call `NewGlobalnetConfigMap` and create CM directly:

```
configMap, err := globalnet.NewGlobalnetConfigMap(brokerObj.Spec.GlobalnetEnabled,
		brokerObj.Spec.GlobalnetCIDRRange, brokerObj.Spec.DefaultGlobalnetClusterSize, brokerNamespace)
if err == nil {
    configMap.Labels[BackupLabel] = "submariner"
    err = client.Create(ctx, c.controllerClient)
}

return return errors.Wrapf(err, "error creating globalnet configmap on Broker")
```

pkg/hub/submarineragent/controller.go
```diff
@@ -687,12 +689,44 @@ func (c *submarinerAgentController) createGNConfigMapIfNecessary(ctx context.Con
...
+		err = c.controllerClient.Create(ctx, configMap)
 	}
 
-	return nil
+	return errors.Wrapf(err, "error creating globalnet configmap on Broker")
+}
+
+func (c *submarinerAgentController) updateBackupLabelOnGnConfigMap(ctx context.Context, namespace string) error {
+	return errors.Wrapf(retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		logger.Infof("Adding backup label to globalnet configmap in %q", namespace)
```
tpantelis, line 704: I think this will cause too much logging, especially since this is only for upgrade. The logging on L724 should suffice.

sridhargaddam: ++. 

Also, how frequently this API will be called?

vthapar: Aha, sorry, this was for debugging/testing and missed to remove it before pushing.

tpantelis (approved):

tpantelis (approved):

nyechiel:
@tpantelis can you please look at backporting this to 2.7 while Vishal is out?

PR #902, opened 2023-05-28
Set Submariner.Spec.Version

tpantelis:


skitt (changes requested):
What’s the goal behind this change?

The addon is supposed to deploy Submariner components as specified by the bundle, using image digests defined there; to support that, we _removed_ the `version` field which used to be hard-coded (#631).

skitt:
> Where are the image digests defined? It's the operator that creates the DaemonSets and when it reads the `Submariner` resource, the `Version` field will again be set to the version baked into the binary, ie `Version` is never empty.

They’re defined [during the downstream build](https://gitlab.cee.redhat.com/acm-cpaas-midstream/submariner-operator-midstream/-/blob/bc8501434143c803dbb16c14a06b5acb4271bd6a/distgit/containers/submariner-operator-bundle/render_templates#L53), and stored in the operator bundle which is published in the operator catalog.

The operator [ignores `Version` if related images are specified in the bundle](https://github.com/submariner-io/submariner-operator/blob/ebdbdf361dbb3ee577491c886f0353a6b2a46597/pkg/images/images.go#L36), so setting `Version` in this way wouldn’t break anything, but it would still have the potential to be misleading.

I suppose we can apply this for now, but we should really add the version information to the status and use that in `subctl`.

tpantelis:
> The operator [ignores Version if related images are specified in the bundle](https://github.com/submariner-io/submariner-operator/blob/ebdbdf361dbb3ee577491c886f0353a6b2a46597/pkg/images/images.go#L36), so setting Version in this way wouldn’t break anything, but it would still have the potential to be misleading.

Why would/could it be misleading? Is `Version` intended to be the image version and only set in the CR if overridden by the user via `subctl join`? If so then why do we default it [here](https://github.com/submariner-io/submariner-operator/blob/029aa298708e358bb8b73ddbd0abb48b6faa95f1/api/v1alpha1/submariner_types.go#L351) when read? I'd think hat wold be misleading as well. 

Perhaps we should change `subctl gather` to not display `Version` if it's misleading.

tpantelis:
> ...we should really add the version information to the status and use that in subctl.

that sounds reasonable.

PR #911, opened 2023-06-01
Obtain broker CRDs from submariner-operator

tpantelis:
...instead of the local embedded YAML files to ensure the most up-to-date CRDs are installed.


sridhargaddam:
Thanks for pushing the fix @tpantelis 

/lgtm

PR #920, opened 2023-06-06
Align linting with Submariner

skitt:
This updates the golangci-lint target to install a version of golangci-lint determined by the project, instead of relying on externally-provided images. It updates golangci-lint to v1.52.2 (not the latest, but matching Submariner) and fixes a few issues.


tpantelis:
I guess this depends on https://github.com/openshift/release/pull/40061#event-9445424464

PR #928, opened 2023-06-08
Allow Submariner debugging to be enabled

skitt:
The addon currently forcibly disables debugging in Submariner. This changes that; debugging Submariner and IPSec can be enabled in SubmarinerConfig.

### DIFF
--- a/pkg/apis/submarinerconfig/v1alpha1/types.go
+++ b/pkg/apis/submarinerconfig/v1alpha1/types.go
@@ -74,6 +74,16 @@ type SubmarinerConfigSpec struct {
 	// +kubebuilder:default=false
 	InsecureBrokerConnection bool `json:"insecureBrokerConnection"`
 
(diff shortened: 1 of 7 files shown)


pkg/apis/submarinerconfig/v1alpha1/types.go
```diff
@@ -74,6 +74,14 @@ type SubmarinerConfigSpec struct {
 	// +kubebuilder:default=false
 	InsecureBrokerConnection bool `json:"insecureBrokerConnection"`
 
+	// IPSecDebug enables IPSec debugging.
+	// +optional
+	IPSecDebug bool `json:"IPSecDebug,omitempty"`
```
sridhargaddam, line 79: should we use here and below?
// +kubebuilder:default=false

sridhargaddam:
@skitt shall we backport this PR to other branches too?

PR #938, opened 2023-06-12
Redact sensitive data in log output

tpantelis:
...as is done in subctl, eg

  `"brokerK8sApiServer": "##redacted##"`


skitt:
> @skitt The logging was introduced in 2.8 so only needs to be cherry-picked to release-2.8.

Ah yes, I can’t cancel the requests but they’ll fail on their own.

PR #965, opened 2023-06-16
Preserve Gateway labels added by the user

sridhargaddam:
For certain platforms like VmWare, Submariner addon labels one (or more) of the existing worker nodes as Gateway nodes when its installed on the ManagedCluster. However, if the clusterAdmin/user explicitly requires certain nodes to be used as Gateway nodes, we recommend that `submariner.io/gateway=true` label has to be added to the desired nodes. Currently, the addon code looks for the presence of `submariner.io/gateway=true` label and uses those nodes accordingly, but when its later uninstalled, it removes the label from the node which was originally added by the user.

This PR ensures that the label is preserved if its not added by the submariner-addon.

### DIFF
--- a/pkg/spoke/submarineragent/gateways_controller.go
+++ b/pkg/spoke/submarineragent/gateways_controller.go
@@ -21,6 +21,7 @@ import (
 
 const (
 	submarinerGatewayLabel        = "submariner.io/gateway"
+	gatewayLabeledBySubmariner    = "submariner.io/random-gateway-node"
 	submarinerGatewayNodesLabeled = "SubmarinerGatewayNodesLabeled"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/spoke/submarineragent/gateways_controller.go
```diff
@@ -21,6 +21,7 @@ import (
 
 const (
 	submarinerGatewayLabel        = "submariner.io/gateway"
+	GatewayLabelledBySubmariner   = "submariner.io/random-gateway-node"
```
skitt, line 24: This can be private.

skitt (changes requested):
Apart from the public/private issue, this seems OK. On upgrades we’ll err on the side of caution: existing gateways will be treated as labelled by the administrator, not the addon.

pkg/spoke/submarineragent/gateways_controller.go
tpantelis: Also one 'l':

```suggestion
	gatewayLabeledBySubmariner   = "submariner.io/random-gateway-node"
```

skitt: That’s a UK/US variation, either is fine by me.

PR #1017, opened 2023-07-18
Set default gateways count to one in submarinerConfig

sridhargaddam:
Currently, the SubmarinerConfigSpec specifies gatewayConfig as an Optional field and because of this, when users create a near empty SubmarinerConfig, the deployment does not create any gateways. This PR removes the Optional field from gateways, so that a default gateway of one is used in such deployments.

### DIFF
--- a/pkg/apis/submarinerconfig/v1alpha1/types.go
+++ b/pkg/apis/submarinerconfig/v1alpha1/types.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"encoding/json"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -194,7 +196,7 @@ type GatewayConfig struct {
 	// Submariner gateway HA will be enabled automatically.
 	// +optional
 	// +kubebuilder:default=1
-	Gateways int `json:"gateways,omitempty"`
+	Gateways int `json:"gateways"`
 }
 
 type AWS struct {
@@ -288,4 +290,22 @@ type SubmarinerConfigList struct {
 
 	// Items is a list of SubmarinerConfig.
 	Items []SubmarinerConfig `json:"items"`
+}
+
+func (s *SubmarinerConfig) UnmarshalJSON(data []byte) error {
+	type SubmarinerConfigAlias SubmarinerConfig
+
+	subm := &SubmarinerConfigAlias{
+		Spec: SubmarinerConfigSpec{
+			GatewayConfig: GatewayConfig{
+				Gateways: 1,
+			},
+		},
+	}
+
+	_ = json.Unmarshal(data, subm)
+
+	*s = SubmarinerConfig(*subm)
+
+	return nil
 }


sridhargaddam:
/CC @skitt @tpantelis 

Note: I could not verify this PR, but I hope that the change should not create any issues.

skitt:
> Note: I could not verify this PR, but I hope that the change should not create any issues.

I suspect it will prevent upgrades on systems with an existing `SubmarinerConfig` with no `gateways` value. I wonder if it wouldn’t be better to provide a default of 1 when unmarshaling...

skitt:
> > I suspect it will prevent upgrades on systems with an existing `SubmarinerConfig` with no `gateways` value.
> 
> Wondering why someone will configure a submarinerConfig with zero gateways.

True, I wouldn’t expect a setup with 0 gateways to be upgraded. But the whole point of this PR is to handle scenarios with unspecified gateways ;-).

> > I wonder if it wouldn’t be better to provide a default of 1 when unmarshaling...
> 
> If we take this approach, how can we distinguish between a SubmarinerConfig that does not have `gatewayConfig` vs `gatewayConfig.gateways` explicitly set to zero by the user?

If it’s set to 0, it will have a value on unmarshaling. The way to deal with it is to set gateways to 1 before unmarshaling, something like

```
func (s *SubmarinerConfig) UnmarshalJSON(data []byte) error {
	type submarinerConfigAlias SubmarinerConfig

	submarinerConfig := &submarinerConfigAlias{
		Spec: SubmarinerConfigSpec{
			GatewayConfig: GatewayConfig{
				Gateways: 1,
			],
		},
	}

	_ = json.Unmarshal(data, submarinerConfig)

	*s = SubmarinerConfig(*submarinerConfig)

	return nil
}
```

Thus if the user explicitly sets gateways to 0, that value will be unmarshaled on top of the default of 1, replacing it.

sridhargaddam:
> Thus if the user explicitly sets gateways to 0, that value will be unmarshaled on top of the default of 1, replacing it.

Thanks for the code snippet, but I'm seeing that even if the user explicitly sets the `gateways` to 0, json.Unmarshal resets it to 1. Am I missing something?

```
package main

import (
    "encoding/json"
    "fmt"
)

type SubmarinerConfig struct {
    Spec SubmarinerConfigSpec `json:"spec"`
}

type SubmarinerConfigSpec struct {
    CableDriver string `json:"cableDriver,omitempty"`
    GatewayConfig `json:"gatewayConfig,omitempty"`
}

type GatewayConfig struct {
    AWS `json:"aws,omitempty"`
    Gateways int `json:"gateways,omitempty"`
}

type AWS struct {
    InstanceType string `json:"instanceType,omitempty"`
}

func main() {
    userSMConfig := &SubmarinerConfig{
        Spec: SubmarinerConfigSpec{
            CableDriver: "vxlan",
            GatewayConfig: GatewayConfig{
                Gateways: 0,
            },
        },
    }

    userSMjsonData, err := json.Marshal(userSMConfig)
    if err != nil {
        fmt.Println("Error:", err)
        return
    }

    smConfig := &SubmarinerConfig{
        Spec: SubmarinerConfigSpec{
            CableDriver: "libreswan",
            GatewayConfig: GatewayConfig{
                Gateways: 1,
            },
        },
    }

    err = json.Unmarshal(userSMjsonData, &smConfig)
    if err != nil {
        fmt.Println("Error:", err)
        return
    }

    fmt.Println("userSMConfig is: ", *userSMConfig)
    fmt.Println("smConfig is: ", *smConfig)
    return
}

```

Output:

```
userSMConfig is:  {{vxlan {{} 0}}}
smConfig is:  {{vxlan {{} 1}}}

```

skitt:
> > Thus if the user explicitly sets gateways to 0, that value will be unmarshaled on top of the default of 1, replacing it.
> 
> Thanks for the code snippet, but I'm seeing that even if the user explicitly sets the `gateways` to 0, json.Unmarshal resets it to 1. Am I missing something?

The issue there is that `Gateways` is marked `omitempty` — 0 is the empty value, so the serialized JSON drops the field entirely. If you remove that marker, everything works as expected.

sridhargaddam:
> The issue there is that `Gateways` is marked `omitempty` — 0 is the empty value, so the serialized JSON drops the field entirely. If you remove that marker, everything works as expected: https://play.golang.com/p/8Qk7ncuayYo

Thanks @skitt. I'm trying to understand the appropriate place where this transformation can be applied and see the following code as a possibility.
https://github.com/stolostron/submariner-addon/blob/ab3efb9268aa5c4a69b075415e4708c93f1bf47c/pkg/hub/submarineragent/controller.go#L240

But this API will be called periodically, so it may not be a good idea to apply it there. Do you have any suggestion on where to apply this change?

skitt:
> Thanks @skitt. I'm trying to understand the appropriate place where this transformation can be applied and see the following code as a possibility.
> 
> https://github.com/stolostron/submariner-addon/blob/ab3efb9268aa5c4a69b075415e4708c93f1bf47c/pkg/hub/submarineragent/controller.go#L240
> 
> But this API will be called periodically, so it may not be a good idea to apply it there. Do you have any suggestion on where to apply this change?

The unmarshaling code should go in the `UnmarshalJSON` attached to the relevant type, in `pkg/apis/submarinerconfig/v1alpha1/types.go`. See `submariner_types.go` in the `submariner` project for an example.

sridhargaddam:
> The unmarshaling code should go in the `UnmarshalJSON` function attached to the relevant type, in `pkg/apis/submarinerconfig/v1alpha1/types.go`. See `submariner_types.go` in the `submariner` project for an example. The code shown in [#1017 (comment)](https://github.com/stolostron/submariner-addon/pull/1017#issuecomment-1642348262) might just work there as-is.

Ack, thanks. Updated the PR.

skitt (approved):

PR #1018, opened 2023-07-18
Allow the addon to patch clustermanagementaddons/status

skitt:
This seems to be required for syncing now.


nyechiel:
Waiting for feedback from @MaxBab if this change resolves the issue he is seeing.

deploy/config/rbac/cluster_role.yaml
```diff
@@ -65,6 +65,9 @@ rules:
 - apiGroups: ["addon.open-cluster-management.io"]
   resources: ["clustermanagementaddons/finalizers"]
   verbs: ["update"]
+- apiGroups: ["addon.open-cluster-management.io"]
+  resources: ["clustermanagementaddons/status"]
+  verbs: ["patch", "update"]
```
sridhargaddam, line 70: Do we also have to update `deploy/olm-catalog/manifests/submariner-addon.clusterserviceversion.yaml`

skitt: Indeed, thanks @sridhargaddam!

skitt:
This doesn’t resolve the main problem @MaxBab is seeing, but it does fix one permission error.

/cherrypick release-2.8

PR #1046, opened 2023-08-03
Implement leader election settings

tpantelis:
...according to https://issues.redhat.com/browse/ACM-1441.


tpantelis:
There is no longer a requirement to make the settings configurable. Since the Openshift controller lib defaults to the default to the recommended settings, this PR isn't needed.

PR #1057, opened 2023-08-07
Update dependabot config

tpantelis:
- Track only submariner-operator, as it includes admiral and submariner, and cloud-prepare.
- Remove config for 2.5 as we're no longer releasing from it

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,23 +17,9 @@ updates:
       - dependency-name: k8s.io/client-go
       - dependency-name: k8s.io/code-generator
       - dependency-name: k8s.io/component-base
-  - package-ecosystem: gomod
-    target-branch: "release-2.5"
-    directory: "/"
-    schedule:
-      interval: weekly
-    open-pull-requests-limit: 5
-    ignore:
-      # K8s and operator SDK, we need to handle these manually
-      - dependency-name: github.com/operator-framework/*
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -48,8 +34,11 @@ updates:
       # Addon framework, this shouldn't be upgraded on release branches
       - dependency-name: open-cluster-management.io/addon-framework
       - dependency-name: open-cluster-management.io/api
+      # These are included with submariner-operator
+      - dependency-name: github.com/submariner-io/admiral
+      - dependency-name: github.com/submariner-io/submariner
       # 2.6 tracks Submariner 0.13
-      - dependency-name: github.com/submariner-io/*
+      - dependency-name: github.com/submariner-io/submariner-operator
```
skitt, line 41: Don’t we need to track `cloud-prepare` too?

tpantelis: yeah

.github/dependabot.yml
```diff
@@ -48,8 +34,13 @@ updates:
...
+        versions: ">= 0.14.0-m0"
+      - dependency-name: github.com/submariner-io/cloud-prepare
```
sridhargaddam, line 43: Do you want to amend the commit message accordingly?

tpantelis: sure

PR #1090, opened 2023-08-31
Increase the linting timeout to 15 minutes

skitt:


skitt:
The failing linting jobs were due to not having enough memory; increasing the timeout isn’t useful.

PR #1121, opened 2023-09-27
Refactor/cleanup integration tests

tpantelis:
General outline below but see individual commits for details.
 
- Refactor common/duplicate code to reusable functions
 - Improve test description and trace text
 - Remove unnecessary code
 - Fix the spoke agent test case
 - Gracefully stop controller manager in `AfterSuite`

### DIFF
--- a/pkg/hub/manager.go
+++ b/pkg/hub/manager.go
@@ -185,12 +185,13 @@ func (o *AddOnOptions) RunControllerManager(ctx context.Context, controllerConte
 		controllerContext.EventRecorder,
 	)
 
-	go clusterInformers.Start(ctx.Done())
-	go workInformers.Start(ctx.Done())
-	go kubeInformers.Start(ctx.Done())
-	go configInformers.Start(ctx.Done())
-	go apiExtensionsInformers.Start(ctx.Done())
-	go addOnInformers.Start(ctx.Done())
+	clusterInformers.Start(ctx.Done())
+	workInformers.Start(ctx.Done())
+	kubeInformers.Start(ctx.Done())
+	configInformers.Start(ctx.Done())
+	apiExtensionsInformers.Start(ctx.Done())
+	addOnInformers.Start(ctx.Done())
+
 	go submarinerBrokerCRDsController.Run(ctx, 1)
 	go submarinerBrokerController.Run(ctx, 1)
 	go submarinerAgentController.Run(ctx, 1)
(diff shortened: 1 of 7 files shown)


tpantelis:
The sonar job fails b/c the integration test code of course has no unit tests.

test/integration/deploying_submariner_test.go
```diff
@@ -322,19 +240,27 @@ var _ = Describe("Deploy a submariner on hub", func() {
 })
 
 func awaitSubmarinerManifestMorks(managedClusterName string) {
```
sridhargaddam, line 242: nit: ...Works

test/integration/deploying_submariner_test.go
```diff
@@ -322,19 +240,27 @@ var _ = Describe("Deploy a submariner on hub", func() {
...
-	By("Check if the submariner agent manifestworks are deployed")
+	By("Await deployment of the ManifestWorks")
 	Eventually(func() bool {
-		return util.FindManifestWorks(workClient, managedClusterName, submarineragent.OperatorManifestWorkName,
+		return util.CheckManifestWorks(workClient, managedClusterName, true, submarineragent.OperatorManifestWorkName,
 			submarineragent.SubmarinerCRManifestWorkName)
 	}, eventuallyTimeout, eventuallyInterval).Should(BeTrue())
 }
 
+func ensureSubmarinerManifestMorks(managedClusterName string) {
```
sridhargaddam, line 250: Same as above

test/integration/deploying_submariner_test.go
```diff
@@ -322,19 +240,27 @@ var _ = Describe("Deploy a submariner on hub", func() {
...
+	By("Ensure deployment of the ManifestWorks")
+	Consistently(func() bool {
+		return util.CheckManifestWorks(workClient, managedClusterName, true, submarineragent.OperatorManifestWorkName,
+			submarineragent.SubmarinerCRManifestWorkName)
+	}, 1).Should(BeTrue())
+}
+
 func awaitNoSubmarinerManifestMorks(managedClusterName string) {
```
sridhargaddam, line 258: ditto

pkg/hub/manager.go
```diff
@@ -185,12 +185,13 @@ func (o *AddOnOptions) RunControllerManager(ctx context.Context, controllerConte
...
-	go configInformers.Start(ctx.Done())
-	go apiExtensionsInformers.Start(ctx.Done())
-	go addOnInformers.Start(ctx.Done())
+	clusterInformers.Start(ctx.Done())
+	workInformers.Start(ctx.Done())
+	kubeInformers.Start(ctx.Done())
+	configInformers.Start(ctx.Done())
+	apiExtensionsInformers.Start(ctx.Done())
+	addOnInformers.Start(ctx.Done())
+
```
sridhargaddam, line 194: The commit message associated with this change does not talk about why we stopped using goroutines. 
Is it because the SharedInformerFactory.Start() method internally uses goroutines? If so, can you include it in the commit message.

tpantelis: > Is it because the SharedInformerFactory.Start() method internally uses goroutines? 

yes. I guess I threw in that change w/o mentioning it :)

PR #1122, opened 2023-09-28
Add rbac to update ClusterManagementAddon resource

tpantelis:
...specifically the finalizer. There is an entry for `update` access to the `clustermanagementaddons/finalizers` resource but this does not work since `finalizers` is not defined as a subresource in the CRD so remove this entry. Same with `managedclusteraddons/finalizers`.


tpantelis:
/cc @skitt @sridhargaddam 

This is needed ASAP.

deploy/config/rbac/cluster_role.yaml
```diff
@@ -61,19 +61,13 @@ rules:
   verbs: ["get", "list", "watch"]
 - apiGroups: ["addon.open-cluster-management.io"]
   resources: ["clustermanagementaddons"]
-  verbs: ["create", "get", "list", "watch"]
-- apiGroups: ["addon.open-cluster-management.io"]
-  resources: ["clustermanagementaddons/finalizers"]
```
sridhargaddam, line 66: Any idea how it worked all these days?

skitt: We never updated finalizers in the past; if I’ve understood correctly, this requirement is related to https://github.com/stolostron/submariner-addon/pull/1120/files#diff-5ee37ee4523e50d6f35fbd53ff29e9caf343e30e3786fac879a7d4167ca3ad8cR192

sridhargaddam: It was added as part of the following change - https://github.com/stolostron/submariner-addon/pull/840

skitt: Ah, I remember discussing this with @vthapar but I can’t remember where we saw the issue in the framework...

tpantelis: From what I read it seems like specifying "*/finalizers" **should** work but when tested in an ACM setup, the controller got a permission error trying to remove the finalizer:

`clustermanagementaddons.addon.open-cluster-management.io "submariner" is forbidden: User "system:serviceaccount:open-cluster-management:submariner-addon" cannot update resource "clustermanagementaddons" in API group "addon.open-cluster-management.io" at the cluster scope`

The MCH has the change from #840 so it's puzzling: https://github.com/stolostron/multiclusterhub-operator/blob/b58ebefd4d7b8ff0082d2950bec0e2bd1860e7fc/pkg/templates/charts/toggle/submariner-addon/templates/submariner-addon-clusterrole.yaml#L184

Anyway the submariner add-on should be able to update anything on its own CMA. We create the CMA on startup of the agent [here](https://github.com/stolostron/submariner-addon/blob/2b51541c563a4161d8ae9e95129097f7f0a12911/pkg/hub/manager.go#L222) but we should do create-or-update to handle upgrades, not just to add the finalizer but in case we change something in the spec.

PR #1123, opened 2023-09-29
Use CreateOrUpdate for the ClusterManagementAddOn 

tpantelis:
...to handle upgrades. This is initially to ensure the finalizer is added but will be useful in the future if the spec is ever changed.

Also added integration test case. See individual commits for details.


skitt:
/approve

(This will need to be force-merged.)

PR #1142, opened 2023-10-06
Cleanup broker resources in empty clusterset

vthapar:
When clusterset is empty, cleanup broker resources. This will allow a new deployment to use fresh settings without needing to delete/re-create the clusterset.

### DIFF
--- a/pkg/hub/submarinerbroker/controller.go
+++ b/pkg/hub/submarinerbroker/controller.go
@@ -174,6 +174,11 @@ func (c *submarinerBrokerController) reconcileManagedClusterSet(ctx context.Cont
 		return c.doClusterSetCleanup(ctx, clusterSet, recorder)
 	}
 
+	if meta.IsStatusConditionPresentAndEqual(clusterSet.Status.Conditions, clusterv1beta2.ManagedClusterSetConditionEmpty,
+		metav1.ConditionTrue) && finalizer.IsPresent(clusterSet, brokerFinalizer) {
+		return c.deleteBrokerResources(ctx, clusterSet, recorder)
+	}
+
 	brokerNS := brokerinfo.GenerateBrokerName(clusterSet.Name)
 
 	if !finalizer.IsPresent(clusterSet, brokerFinalizer) || clusterSet.GetAnnotations()[SubmBrokerNamespaceKey] != brokerNS {
@@ -236,6 +241,31 @@ func (c *submarinerBrokerController) createIPSecPSKSecret(ctx context.Context, b
 	}
 
 	return err
+}
+
+func (c *submarinerBrokerController) deleteBrokerResources(ctx context.Context, clusterSet *clusterv1beta2.ManagedClusterSet,
+	recorder events.Recorder,
+) error {
+	brokerNS := clusterSet.GetAnnotations()[SubmBrokerNamespaceKey]
+	if brokerNS == "" {
+		return nil
+	}
+	// TODO: Ideally, submariner finalizer check in caller should be enough and
+	//       we should just delete resources unconditionally if here. But we need
+	//       this check coz we're adding finalizer to non-submariner clustersets too
+	_, err := c.kubeClient.CoreV1().Namespaces().Get(ctx, brokerNS, metav1.GetOptions{})
+
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+
+	if err != nil {
+		return errors.Wrapf(err, "error getting namespace %q", brokerNS)
+	}
+
+	logger.Infof("Deleting broker resources for  %q", brokerNS)
+
+	return resource.DeleteFromManifests(ctx, c.kubeClient, recorder, assetFunc(brokerNS), staticResourceFiles...)
 }
 
 func (c *submarinerBrokerController) doClusterSetCleanup(ctx context.Context, clusterSet *clusterv1beta2.ManagedClusterSet,
--- a/pkg/hub/submarinerbroker/controller.go
+++ b/pkg/hub/submarinerbroker/controller.go
@@ -15,6 +15,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/finalizer"
 	"github.com/submariner-io/admiral/pkg/log"
 	coreresource "github.com/submariner-io/admiral/pkg/resource"
+	"github.com/submariner-io/admiral/pkg/slices"
 	"github.com/submariner-io/admiral/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/hub/submarinerbroker/controller.go
```diff
@@ -238,6 +242,41 @@ func (c *submarinerBrokerController) createIPSecPSKSecret(ctx context.Context, b
...
+func (c *submarinerBrokerController) deleteBrokerResources(ctx context.Context, clusterSet *clusterv1beta2.ManagedClusterSet,
+	recorder events.Recorder,
+) error {
+	brokerNS := clusterSet.GetAnnotations()[SubmBrokerNamespaceKey]
+	if brokerNS == "" {
+		return nil
+	}
+	// TODO: Ideally, submariner finalizer check in caller should be enough and
+	//       we should just delete resources unconditionally if here. But we need
+	//       this check coz we're adding finalizer to non-submariner clustersets too
```
tpantelis, line 264: I'm not sure I follow this. What does the finalizer have to do with deleting the broker NS? `resource.DeleteFromManifests` handles deleting non-existent resource.

pkg/hub/submarinerbroker/controller.go
```diff
@@ -174,6 +174,10 @@ func (c *submarinerBrokerController) reconcileManagedClusterSet(ctx context.Cont
 		return c.doClusterSetCleanup(ctx, clusterSet, recorder)
 	}
 
+	if isClusterSetEmpty(clusterSet) && finalizer.IsPresent(clusterSet, brokerFinalizer) {
```
tpantelis, line 177:
```suggestion
	if meta.IsStatusConditionPresentAndEqual(clusterSet.Status.Conditions, clusterv1beta2.ManagedClusterSetConditionEmpty, metav1.ConditionTrue) && finalizer.IsPresent(clusterSet, brokerFinalizer) {
```

pkg/hub/submarinerbroker/controller.go
```diff
@@ -174,6 +174,10 @@ func (c *submarinerBrokerController) reconcileManagedClusterSet(ctx context.Cont
...
+		return c.deleteBrokerResources(ctx, clusterSet, recorder)
+	}
```
tpantelis, line 179: When an `ManagedClusterSet` (MCS) is first created it won't have any `ManagedClusters` (MC). So after we add the finalizer below, it returns and this will prevent subsequent updates from applying the broker resources until the MCS is no longer empty. This seems fine.

It would be useful to add unit tests for this flow. Since I'm familiar with the tests I can add them if you want.

tpantelis (changes requested):

pkg/hub/submarinerbroker/controller.go
vthapar: Idea is to use presence of finalizer as a check for submariner being installed on this clusterset and requiring cleanup or not. If we were to add finalizer only when installing submarineraddon, this check would be sufficent without needing to get manifestworks, namespace or anything else.

pkg/hub/submarinerbroker/controller.go
vthapar: currently finalizer is added on clustersets, even when they are empty and submariner is not installed. e.g. `default` clusterset. A better check would allow us to add a `Info` log here without it getting spammed by other empty clustersets.

pkg/hub/submarinerbroker/controller.go
tpantelis: I still don't see why you need to explicitly check if the `Namespace` exists before calling `resource.DeleteFromManifests` since `resource.DeleteFromManifests` handles `IsNotFound` error anyway.

pkg/hub/submarinerbroker/controller.go
```diff
@@ -174,6 +174,11 @@ func (c *submarinerBrokerController) reconcileManagedClusterSet(ctx context.Cont
 		return c.doClusterSetCleanup(ctx, clusterSet, recorder)
 	}
 
+	if meta.IsStatusConditionPresentAndEqual(clusterSet.Status.Conditions, clusterv1beta2.ManagedClusterSetConditionEmpty,
```
sridhargaddam, line 177: Does this API (reconcileManagedClusterSet) get invoked periodically? 

If so, can we think of a way to optimize this check as otherwise the condition (clusterv1beta2.ManagedClusterSetConditionEmpty) would always be true for an empty clusterSet and we will be invoking deleteBrokerResources() at regular intervals.

tpantelis: It should only get invoked when something changes in the `ManagedClusterSet`.

sridhargaddam: Okay, then its good.

vthapar: Yes, this is invoked periodically. That is why I have an extra check for finalizer. and didn't add any log here. There is also extra check in `deleteBrokerResources` that checks of broker namespace is present or not. Ideally finalizer+empty check should be enough but since we add finalizer even for clustersets without submariner, it doesn't help much yet.

sridhargaddam: Ack. Can you point me to the code where we create the brokerNamespace only when a ManagedCluster is added to the clusterSet?

tpantelis: @sridhargaddam We don't, ie we create the brokerNamespace when a ManagedClusterSet is created.  But only ManagedClusterSets of type ExclusiveClusterSetLabel.

tpantelis: @vthapar What periodic update cadence are you seeing? I ran the integration tests and, after steady state, saw a `ManagedClusterSet` update after 10 min.

So the extra `Get` of the broker namespace in `deleteBrokerResources` is intended as a shortcut to reduce API calls due to periodic update. `DeleteFromManifests` does a single `Delete` but it attempts to delete 2 resources so the extra `Get` does save one API call.

But there's another simple check we could do. We combine adding the finalizer and the namespace label in [one Update operation](https://github.com/stolostron/submariner-addon/blob/a5cd8c42a701b16edd7566206cf9328d9c4d6d32/pkg/hub/submarinerbroker/controller.go#L179). But if we add the finalizer first and return when added, then we can also check  for the presence of the label:

```
if meta.IsStatusConditionPresentAndEqual(...) && finalizer.IsPresent(...) && 
            clusterSet.GetAnnotations()[SubmBrokerNamespaceKey] != "" {
    ....
}
```

vthapar: @tpantelis Sorry, I was not clear enough. You're right, it is not periodic but on clusteset chanes. But with this code if I only check for finalizer and empty to clean up, it was hitting repeatedly for `default` clusterset. For some reason we add finalizer even to default clusterset even when there is no managedcluster added to it, and that is what was causing periodic updates till I added check for namespace in `deleteBrokerResources`

Not sure I get comment about returning after adding finalizer. I am doing get on Namespace coz for `default` clusterset which had no clusters or submariner installed, I didn't see the namespace though the annotation and finalizer still exists. I added the Namespace check in called method so that in future when we only add finalizer to clustersets which have submariner installed, we can skip namespace Get in `deleteBrokerResources` and just use finalizer as required check.

tpantelis: I do see a periodic update cadence of 10 min.

I don't understand why we get repeated updates w/o the namespace check. `DeleteFromManifests` just does a delete on the 2 broker resources. It doesn't make sense why that would cause repeated updates for the MCS. We need to understand that. I assume you had not added any `ManagedCluster` to the default MCS. I'll pull in your PR and try it with the integration tests.

When we add the finalizer and/or namespace annotation, if either was actually added we return immediately and don't apply the broker resources yet. We wait for the next update to do so. This has the side effect such that with your changes, we don't apply the broker resources if the MCS is empty. By separating adding the finalizer from the namespace annotation, I think we can check for presence of the namespace annotation instead of the namespace resource.

We really need unit tests to ensure correctness. I can pull in your PR and add tests.

tpantelis: > due to deleting resources on default clusterset [including delete finalizer]

But you're not deleting the finalizer in  your new code.

> And won't adding a finalizer to clusterset trigger an event for subsequent update operation?

yes

vthapar:
During testing and discussions with Tom we noticed that there is similar issue with cleaning up broker resources when submariner addon is removed from the last cluster in clusterset, but clusterset is not empty. 

https://github.com/stolostron/submariner-addon/pull/1155 should fix both issues, so we will go with 1155 as a more complete and better fix.

sridhargaddam (changes requested):
@vthapar Based on your [last comment](https://github.com/stolostron/submariner-addon/pull/1142#issuecomment-1758892296), I thought we can close the issue to avoid confusion, but I'm not seeing a close option. So, requesting changes just for other reviewers to know that there is a [separate PR](https://github.com/stolostron/submariner-addon/pull/1155) to address this use-case.

PR #1155, opened 2023-10-11
Delete GN ConfigMap and Broker when last add-on is removed

tpantelis:
...for a `ManagedClusterSet`. This will allow a user to re-configure Globalnet settings w/o re-creating the `ManagedClusterSet`.


vthapar (approved):
Looks good. Pulled locally and tested to verify all scenarios work fine.

PR #1178, opened 2023-10-23
Log a warning if the channel doesn't match the default

skitt:
Users can configure a non-default channel, but support needs to know about this, so log it.


skitt:
@tpantelis is there a fake logger we could use to test this?

pkg/hub/submarinerbrokerinfo/brokerinfo.go
```diff
@@ -214,6 +214,10 @@ func applySubmarinerConfig(brokerInfo *SubmarinerBrokerInfo, submarinerConfig *c
 	setIfValueNotDefault(&brokerInfo.CatalogStartingCSV, submarinerConfig.Spec.SubscriptionConfig.StartingCSV)
 	setIfValueNotDefault(&brokerInfo.InstallPlanApproval, submarinerConfig.Spec.SubscriptionConfig.InstallPlanApproval)
 
+	if brokerInfo.CatalogChannel != defaultCatalogChannel {
+		logger.Warningf("Tracking non-default catalog channel %q (default is %q)", brokerInfo.CatalogChannel, defaultCatalogChannel)
```
tpantelis, line 218: Why is this a warning? Is this something a user isn't supposed to do?

skitt: In most cases they shouldn’t, and in all cases support needs to know about it. It could be an info-level message but then it wouldn’t necessarily be logged...

tpantelis:
> @tpantelis is there a fake logger we could use to test this?

Not that I know of. But I don't think we should verify specific log messages in unit tests. If the visibility of this message is significant enough,  perhaps you should use the event `Recorder` instead. There is an `InMemoryRecorder` in the openshift library that is intended for unit tests where you can query the events and verify.

pkg/hub/submarinerbrokerinfo/brokerinfo.go
tpantelis: If you use `logger.V(...).Infof` it may not be but if you use `logger.Infof`, it should always be logged.

pkg/hub/submarinerbrokerinfo/brokerinfo.go
```diff
@@ -214,6 +214,10 @@ func applySubmarinerConfig(brokerInfo *SubmarinerBrokerInfo, submarinerConfig *c
 	setIfValueNotDefault(&brokerInfo.CatalogStartingCSV, submarinerConfig.Spec.SubscriptionConfig.StartingCSV)
 	setIfValueNotDefault(&brokerInfo.InstallPlanApproval, submarinerConfig.Spec.SubscriptionConfig.InstallPlanApproval)
 
+	if brokerInfo.CatalogChannel != defaultCatalogChannel {
+		logger.Infof("Tracking non-default catalog channel %q (default is %q)", brokerInfo.CatalogChannel, defaultCatalogChannel)
```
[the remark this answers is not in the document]
skitt: Hah I was just checking how often the loop runs...

When we discussed this a while back we decided to log a message for two reasons:
* end-users might notice it if things go wrong, and figure out they need to fix something on their own (assuming they remember they set up a specific channel)
* the log message can provide all the information support needs — both the configured value, and the expected default

I’ll memoize the channel value and only log it the first time it changes.

tpantelis (approved):

PR #1228, opened 2023-11-24
On RHOSP, disable verification if requested

skitt:
The current cloud setup on RHOSP ignores the "verify" setting in clouds.yaml. On setups with self-signed certificates, this means that the addon can't connect to the RHOSP API endpoint.

When "verify" is set to false, this sets up a custom TLS configuration with disabled verification.

Ideally a complete RHOSP implementation would also consider the certificate information which can be specified in clouds.yaml; see PrepareTLSConfig in
https://github.com/gophercloud/utils/blob/master/internal/util.go for details. It would be necessary to determine a way for the files to be provided (in the same way as clouds.yaml).

github.com/gophercloud/utils also supports this, but only through NewServiceClient, not the general-purpose AuthenticatedClient function. Using NewServiceClient would require reworking cloud-prepare.

### DIFF
--- a/pkg/cloud/rhos/rhos.go
+++ b/pkg/cloud/rhos/rhos.go
@@ -1,8 +1,10 @@
 package rhos
 
 import (
+	"crypto/tls"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
@@ -20,6 +22,7 @@ import (
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


skitt:
/cc @aswinsuryan @vthapar 

This is untested, it would be good to check that OSP deployments are at least not broken by this, and ideally that insecure connections can actually be used.

aswinsuryan:
> /cc @aswinsuryan @vthapar
> 
> This is untested, it would be good to check that OSP deployments are at least not broken by this, and ideally that insecure connections can actually be used.

facing some issue wiht ins

> /cc @aswinsuryan @vthapar
> 
> This is untested, it would be good to check that OSP deployments are at least not broken by this, and ideally that insecure connections can actually be used.

I will try testing it and will update

aswinsuryan:
> > /cc @aswinsuryan @vthapar
> > This is untested, it would be good to check that OSP deployments are at least not broken by this, and ideally that insecure connections can actually be used.
> 
> facing some issue wiht ins
> 
> > /cc @aswinsuryan @vthapar
> > This is untested, it would be good to check that OSP deployments are at least not broken by this, and ideally that insecure connections can actually be used.
> 
> I will try testing it and will update

I was not able to bring up the ACM UI in RHOS setup, if I remember correctly last time we used a setup shared by test team, So requested them.

vthapar (approved):

aswinsuryan (approved):

PR #1310, opened 2024-02-06
Update otelgrpc@v0.46.0 to fix CVE-2023-47108

dfarrell07:
I haven't been able to get this opentelemetry dependency tree fixed yet.

Starting with:

go get -u go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.46.0

I get the typical (for otel) errors that are typically resolved by also manually updating the otel dependency before the transition to k/k dependencies.

I've made it this far:

go get go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp go get go.opentelemetry.io/otel/sdk/resource
go get go.opentelemetry.io/otel/exporters/otlp/otlptrace go get go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp go get go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc

But now all the errors only list packages that also give errors.


tpantelis:
I assume you're running into the same issue I ran into - see https://github.com/stolostron/submariner-addon/pull/1263#issuecomment-1881357508. We likely stuck with the current `otel` version unless we can upgrade other major dependencies (at least `openshift/library-go` and `K8s`).

dfarrell07:
Thanks for clarifying @tpantelis! That was a bit of a complex mess.

PR #1316, opened 2024-02-12
Reduce sharedInformer memory usage

vthapar:
Trim ManagedFields to reduce shared informer memory usage. This requires informers generated with  newer version of k8s/client-go [0.29] so currently it is done for following informers:
     - k8s
     - submarinerconfig

Rest informers should be updated once WithTransform() is available.

Refer: https://github.com/kubernetes/kubernetes/pull/118455

DependsOn #1318


skitt:
https://github.com/open-cluster-management-io/api/pull/320 updates the OCM informers.

tpantelis:
> [open-cluster-management-io/api#320](https://github.com/open-cluster-management-io/api/pull/320) updates the OCM informers.

Cool. I updated the code to use it. The work API needs to be updated as well but we can do that later. This PR gets most of them.

vthapar:
@tpantelis We need to override the sonarcheck for this.

PR #1317, opened 2024-02-12
Add dependabot config for release 2.10

vthapar:
Add dependabot config for release 2.10


vthapar:
Now that we are tagging m0 much earlier, do we need to revisit this?

PR #1365, opened 2024-03-28
Update protobuf to 1.33.0 for GO-2024-2611

dfarrell07:


tpantelis:
We can't bump to this version as it causes an incompatibility issue.

skitt:
In a little more detail, a quick poke at this suggests that bumping protobuf would require bumping the Google client APIs, but that runs into issues around OpenTelemetry and compatibility with the old k8s libraries used for 2.7.

PR #1373, opened 2024-04-03
Create secret for SA if not auto-created

### LINKED ISSUE #1372: Submariner 2.10 installation fails on VMWare with OCP 4.15
SubmarinerAddon installation on VMWare running OCP 4.15+ fails. Only submariner-addon pod is created. Addon pod shows error as such:

```
E0321 00:58:14.099083 1 base_controller.go:266] "SubmarinerAgentController" controller failed to sync "perf2", err: failed to create submariner brokerInfo of cluster perf2 : secrets "perf2" not found
```

This is because 4.15+ onwards OCP doesn't create the secret automatically for the SA. Submariner requires this secret to extract broker credentials to configure submariner on managed clusters.

Fix is for addon to create the `Secret` of type `ServiceAccount` if not already present.

Workaround
-------------------
Example workaround steps for Managed cluster `perf2` that is part of clusterset `submariner`

1. Go to Hub cluster.

2. oc apply the following template

 

```
apiVersion: v1
kind: Secret
metadata:
  name: perf2-dockercfg
  namespace: submariner-broker
  annotations:
    kubernetes.io/service-account.name: perf2
    openshift.io/token-secret.name: perf2-dockercfg
    openshift.io/token-secret.value: perf2-dockercfg
type: kubernetes.io/service-account-token
```
 

3. Get the token from the created secret

```
% oc get secrets -n submariner-broker perf2-dockercfg -o yaml 
data:
  token: <token1>
```
4. Decode the data.token to get the real token

`echo <token1 copied from data.token> |base64 -d `
5. Update the token to the created secret annotation

```
% oc edit secrets -n submariner-broker perf2-dockercfg
metadata:
  annotations:
    openshift.io/token-secret.value: <paste the decoded token>
```
6. Link the modified secret to the SA

```
% oc edit sa -n submariner-broker perf2
....
secrets:
- name: perf2-dockercfg
```
7. Check that ManifestWork `submariner-resource` is created in `perf2` namespace. This can take a while as Addon keeps retrying with an exponential backoff. If taking too long, can restart hub addon pod, but recommended to do so only after secrets have been created and linked for all managedclusters.

vthapar:
If there is no secreted created for ManagedCluster SA in broker namespace, create one. This is required to generate credentials for ManagedClusters to access Broker K8s API server.

Fixes #1372

### DIFF
--- a/pkg/hub/submarinerbrokerinfo/brokerinfo.go
+++ b/pkg/hub/submarinerbrokerinfo/brokerinfo.go
@@ -351,6 +351,7 @@ func getBrokerTokenAndCA(ctx context.Context, kubeClient kubernetes.Interface, d
 
 func getTokenSecretForSA(ctx context.Context, client kubernetes.Interface, sa *corev1.ServiceAccount,
 ) (*corev1.Secret, error) {
+	var secret *corev1.Secret
 	saSecrets, err := client.CoreV1().Secrets(sa.Namespace).List(ctx, metav1.ListOptions{
 		FieldSelector: fields.OneTermEqualSelector("type", string(corev1.SecretTypeServiceAccountToken)).String(),
 	})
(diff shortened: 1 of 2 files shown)


pkg/hub/submarinerbrokerinfo/brokerinfo.go
```diff
@@ -386,13 +405,13 @@ func getTokenAndCAFromSecret(ctx context.Context, tokenSecret *corev1.Secret, ku
 	return string(tokenSecret.Data["token"]), base64.StdEncoding.EncodeToString(kubeAPIServerCA), nil
 }
 
-func GenerateBrokerName(clusterSetName string) string {
-	name := fmt.Sprintf("%s-%s", clusterSetName, brokerSuffix)
-	if len(name) > namespaceMaxLength {
-		truncatedClusterSetName := clusterSetName[(len(brokerSuffix) - 1):]
+func GenerateBrokerName(name string) string {
```
skitt, line 408: Why is this changing?

pkg/hub/submarinerbrokerinfo/brokerinfo.go
```diff
@@ -360,19 +361,37 @@ func getTokenSecretForSA(ctx context.Context, client kubernetes.Interface, sa *c
...
+		secret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        GenerateBrokerName(sa.Name),
+				Annotations: map[string]string{corev1.ServiceAccountNameKey: sa.Name},
+			},
+			Type: corev1.SecretTypeServiceAccountToken,
+		}
+
+		secret, err = client.CoreV1().Secrets(sa.Namespace).Create(ctx, secret, metav1.CreateOptions{})
+		if err != nil && !apierrors.IsAlreadyExists(err) {
```
skitt, line 379: If the result is an “already exists” error, this will end up returning an empty secret, AFAICT.

skitt (changes requested):

pkg/hub/submarinerbrokerinfo/brokerinfo.go
vthapar: Just realized that only reason we will hit this error is if someone created similar name secret manually with a different type. In that case we should just return with error and retry later. Earlier I was using a separate `returnSecret` but changed in refactoring which modfied the logic.

pkg/hub/submarinerbrokerinfo/brokerinfo.go
vthapar: The passed argument was named `clusterSetName` Now we're reusing it to also generate similar name with `clusterName` as an argument. So just changed the argument name to make it more generic and remove assumption that it will always be a clusterSetName.

skitt (approved):

PR #1386, opened 2024-04-15
Add delete secrets permission for addon

vthapar:
Add delete secrets permission for submariner-addon clusterrole so it can delete broker secrets created by it.


tpantelis:
The verify job failed - looks like you need to commit changes to _deploy/olm-catalog/manifests/submariner-addon.clusterserviceversion.yaml_.

PR #1403, opened 2024-04-29
Update Dockerfile references for Konflux

dfarrell07:
Relates-to: https://issues.redhat.com/browse/ACM-10822


skitt:
I suspect this won’t make any difference until the Konflux setup is copied from the 2.10 branch...

dfarrell07:
> I suspect this won’t make any difference until the Konflux setup is copied from the 2.10 branch...

I should send it to 2.10 first in this case, it seems.

dfarrell07:
> > I suspect this won’t make any difference until the Konflux setup is copied from the 2.10 branch...
> 
> I should send it to 2.10 first in this case, it seems.

https://github.com/stolostron/submariner-addon/pull/1404

PR #1446, opened 2024-06-14
Add ForceUDPEncaps to submarinerconfig

vthapar:
Add ForceUDPEncaps field to SubmarinerConfig

### DIFF
--- a/pkg/apis/submarinerconfig/v1alpha1/types.go
+++ b/pkg/apis/submarinerconfig/v1alpha1/types.go
@@ -86,6 +86,11 @@ type SubmarinerConfigSpec struct {
 	// +kubebuilder:default=false
 	IPSecDebug bool `json:"IPSecDebug,omitempty"`
 
+	// ForceUDPEncaps forces UDP Encapsulation for IPSec.
+	// +optional
+	// +kubebuilder:default=false
+	ForceUDPEncaps bool `json:"forceUDPEncaps,omitempty"`
+
 	// Debug enables Submariner debugging (in the logs).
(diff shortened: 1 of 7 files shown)


pkg/apis/submarinerconfig/v1alpha1/types.go
```diff
@@ -86,6 +86,11 @@ type SubmarinerConfigSpec struct {
 	// +kubebuilder:default=false
 	IPSecDebug bool `json:"IPSecDebug,omitempty"`
 
+	// ForceUDPEncaps forces UDP Encapsulation for IPSec.
+	// +optional
+	// +kubebuilder:default=false
+	ForceUDPEncaps bool `json:"ForceUDPEncaps,omitempty"`
```
tpantelis, line 92:
```suggestion
	ForceUDPEncaps bool `json:"forceUDPEncaps,omitempty"`
```

tpantelis (changes requested):

yboaron:
Overall code lgtm, and I guess we should be fine with upgrading ACM/Submariner too b/c ForceUDPEncaps defaults to false.


ForceUDPEncaps is relevant only if clusters are reachable via private IPs and ESP traffic is blocked on firewall, so I wonder how we should expose this capability to users. I don't think we need to expose this in the ACM UI.


In addition, I performed fufu test on KIND where ForceUDPEncaps was only defined on single cluster and tunnel was in error state, which probably means that ForceUDPEncaps needs to be defined in all the clusters in clusterset.

PR #1454, opened 2024-06-17
Fix codegen using new k8s functions

skitt:
Commit f3f00bf ("Use K8s kube_codegen.sh instaed of deprecated generate-groups.sh") broke client code generation; a similar commit had previously been reverted by commit bc6bdb1 ("Revert "Use kube_codegen.sh instead of generate-groups.sh""), but that fix is no longer viable because generate-groups.sh is entirely disabled in 1.30.

Running kube_codegen.sh has no effect now, it only defines functions without running them. This adjusts hack/update-codegen.sh to use those functions to generate clients. It also updates the verification codepath to actually verify code generation; if code generation fails to do anything in the future, this will now be detected.

Depends on #1453

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -7,16 +7,28 @@ set -o pipefail
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
 CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../../../k8s.io/code-generator)}
 
-verify="${VERIFY:-}"
+if [ "${VERIFY:-}" = "--verify-only" ]; then
+  outprefix=$(mktemp -d -p .)/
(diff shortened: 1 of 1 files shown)


vthapar (approved):

tpantelis:
The verify job fails: 

```
kube::codegen::gen_client --output-dir ./tmp.etWzvg4EvR/pkg/client/submarinerconfig --output-pkg github.com/stolostron/submariner-addon/pkg/client/submarinerconfig --boilerplate hack/../hack/../hack/boilerplate.txt --with-watch --one-input-api submarinerconfig pkg/apis

unknown argument: --output-dir
```

skitt:
> The verify job fails:
> 
> ```
> kube::codegen::gen_client --output-dir ./tmp.etWzvg4EvR/pkg/client/submarinerconfig --output-pkg github.com/stolostron/submariner-addon/pkg/client/submarinerconfig --boilerplate hack/../hack/../hack/boilerplate.txt --with-watch --one-input-api submarinerconfig pkg/apis
> 
> unknown argument: --output-dir
> ```

Yes, see the dependency in the PR description.

PR #1404, opened 2024-04-29
Add Dockerfile for Konflux, use by default

dfarrell07:
Explicitly don't use vendoring or Konflux will fail on the ambiguity.

Relates-to: https://issues.redhat.com/browse/ACM-10822

PR #1457, opened 2024-06-18
Copy the Konflux container file and setup from the 2.10 branch

skitt:
See individual commits for details.

### DIFF
--- a/Dockerfile.konflux
+++ b/Dockerfile.konflux
@@ -2,7 +2,7 @@ FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_1.21 AS bui
 WORKDIR /go/src/github.com/stolostron/submariner-addon
 COPY . .
 ENV GO_PACKAGE github.com/stolostron/submariner-addon
-RUN make build --warn-undefined-variables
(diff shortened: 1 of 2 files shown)


Dockerfile.konflux
```diff
@@ -0,0 +1,9 @@
+FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_1.21 AS builder
+WORKDIR /go/src/github.com/stolostron/submariner-addon
+COPY . .
+ENV GO_PACKAGE github.com/stolostron/submariner-addon
+RUN make GO_BUILD_FLAGS=-mod=mod build --warn-undefined-variables
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
```
vthapar, line 7: Shouldn't we set this to els image, or is that not ready yet?

vthapar (changes requested):

Dockerfile.konflux
skitt: One thing at a time :wink: https://github.com/stolostron/submariner-addon/pull/1404#discussion_r1634337512

vthapar (approved):

PR #1652, opened 2024-09-25
Add support for AWS custom VPC

aswinsuryan:
Add support for AWS custom VPC

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -129,6 +129,8 @@ func (f *providerFactory) Get(config *configv1alpha1.SubmarinerConfig, eventsRec
 		return nil, true, err
 	}
 
+	info.SubmarinerConfigAnnotations = config.Annotations
+
 	instance, err := providerFn(info)
 
 	return instance, true, err
--- a/pkg/cloud/aws/aws.go
+++ b/pkg/cloud/aws/aws.go
@@ -70,7 +70,34 @@ func NewProvider(info *provider.Info) (*awsProvider, error) {
 		return nil, err
 	}
 
-	cloudPrepare := cpaws.NewCloud(awsClient, info.InfraID, info.Region)
+	var cloudOptions []cpaws.CloudOption
+
+	if info.SubmarinerConfigAnnotations != nil {
+		annotations := info.SubmarinerConfigAnnotations
+
+		if vpcID, exists := annotations["submariner.io/vpc-id"]; exists {
+			cloudOptions = append(cloudOptions, cpaws.WithVPCName(vpcID))
+		}
+
+		if subnetIDList, exists := annotations["submariner.io/subnet-id-list"]; exists {
+			subnetIDs := strings.Split(subnetIDList, ",")
+			for i := range subnetIDs {
+				subnetIDs[i] = strings.TrimSpace(subnetIDs[i])
+			}
+
+			cloudOptions = append(cloudOptions, cpaws.WithPublicSubnetList(subnetIDs))
+		}
+
+		if controlPlaneSGID, exists := annotations["submariner.io/control-plane-sg-id"]; exists {
+			cloudOptions = append(cloudOptions, cpaws.WithControlPlaneSecurityGroup(controlPlaneSGID))
+		}
+
+		if workerSGID, exists := annotations["submariner.io/worker-sg-id"]; exists {
+			cloudOptions = append(cloudOptions, cpaws.WithWorkerSecurityGroup(workerSGID))
+		}
+	}
+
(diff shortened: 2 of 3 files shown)


pkg/cloud/cloud.go
```diff
@@ -124,6 +126,27 @@ func (f *providerFactory) Get(config *configv1alpha1.SubmarinerConfig, eventsRec
...
+			}
+			info.CloudOptions = append(info.CloudOptions, cpaws.WithPublicSubnetList(subnetIDs))
+		}
+		if controlPlaneSGID, exists := annotations["submariner.io/control-plane-sg-id"]; exists {
+			info.CloudOptions = append(info.CloudOptions, cpaws.WithControlPlaneSecurityGroup(controlPlaneSGID))
+		}
+		if workerSGID, exists := annotations["submariner.io/worker-sg-id"]; exists {
+			info.CloudOptions = append(info.CloudOptions, cpaws.WithWorkerSecurityGroup(workerSGID))
+		}
+	}
```
tpantelis, line 148: This really belongs in the AWS cloud code. I think we should store the `SubmarinerConfig` in the `provider.Info` struct rather than the AWS-specific `CloudOptions`.

tpantelis (changes requested):

tpantelis:
Unfortunately we can't bump submariner dependencies to 0.19 yet due to the K8s bump to 0.31. The `github.com/openshift/library-go` needs to bump first, which is pending with https://github.com/openshift/library-go/pull/1771.

```
# github.com/openshift/library-go/pkg/serviceability
/go/pkg/mod/github.com/openshift/library-go@v0.0.0-20240124134907-4dfbf6bc7b11/pkg/serviceability/panic.go:54:65: cannot use crashOnDelay(delayDuration, delayDurationString) (value of type func(interface{})) as func(context.Context, interface{}) value in argument to append
# k8s.io/component-base/metrics/testutil
/go/pkg/mod/k8s.io/component-base@v0.31.0/metrics/testutil/metrics.go:73:59: undefined: expfmt.NewFormat
```

tpantelis:
This error:

```
# k8s.io/component-base/metrics/testutil
/go/pkg/mod/k8s.io/component-base@v0.31.0/metrics/testutil/metrics.go:73:59: undefined: expfmt.NewFormat
```
 
is b/c we're pinning `github.com/prometheus/common` to v0.47.0 but k8s.io/component-base needs v0.55.0. We should be able to remove the pin now.

pkg/cloud/cloud.go
aswinsuryan: @tpantelis Shouldn't we be moving to SubmarinerConfig variables inside submariner config itself , is this ok to have as an interim fix?

aswinsuryan:
> Unfortunately we can't bump submariner dependencies to 0.19 yet due to the K8s bump to 0.31. The `github.com/openshift/library-go` needs to bump first, which is pending with [openshift/library-go#1771](https://github.com/openshift/library-go/pull/1771).
> 
> ```
> # github.com/openshift/library-go/pkg/serviceability
> /go/pkg/mod/github.com/openshift/library-go@v0.0.0-20240124134907-4dfbf6bc7b11/pkg/serviceability/panic.go:54:65: cannot use crashOnDelay(delayDuration, delayDurationString) (value of type func(interface{})) as func(context.Context, interface{}) value in argument to append
> # k8s.io/component-base/metrics/testutil
> /go/pkg/mod/k8s.io/component-base@v0.31.0/metrics/testutil/metrics.go:73:59: undefined: expfmt.NewFormat
> ```

so we ill have to try get this change in just for 0.16?

pkg/cloud/cloud.go
vthapar: @aswinsuryan Cloud specific variables don't make sense in Submariner CR as submariner CR doesn't do anything with cloud-prepare.

We may have scenarios where we need similar CloudOptions for other clouds too. For now we can leave this here, but can leave a TODO to refactor it later.

tpantelis: I’d rather not leave a TODO - in my experience they rarely get addressed later. It seems like we’re taking about 2 different things here. One is how the user specifies the cloud options - right now it’s via annotations on the SC CR. It could be via fields. I’m not sure how else we’d do it. The fact is whatever we do here likely won’t be changed later as that would require migration, docs change etc. So let’s decide the best way now - annotations seem reasonable and the last east invasive. The second is how we handle it in code. I’m merely the suggesting we move the parsing into the AWS code where it belongs by passing the SC CR to the provider instance.  This is simple to do. I can make the change.

aswinsuryan: Done.

tpantelis (approved):

PR #1660, opened 2024-09-27
Add option to skip cloud prepare

tpantelis:
It may be necessary to skip cloud prepare in some installations, eg to temporarily work around an issue. Introduce a backdoor way to do this via an annotation on the `SubmarinerConfig`.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -3,6 +3,7 @@ package cloud
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/openshift/library-go/pkg/operator/events"
 	configv1alpha1 "github.com/stolostron/submariner-addon/pkg/apis/submarinerconfig/v1alpha1"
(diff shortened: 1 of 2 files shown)


aswinsuryan (approved):

pkg/cloud/cloud.go
```diff
@@ -77,6 +78,10 @@ func NewProviderFactory(restMapper meta.RESTMapper, kubeClient kubernetes.Interf
 }
 
 func (f *providerFactory) Get(config *configv1alpha1.SubmarinerConfig, eventsRecorder events.Recorder) (Provider, bool, error) {
+	if config.Annotations["submariner.io/skip-cloud-prepare"] == strconv.FormatBool(true) {
```
skitt, line 81:
```suggestion
	if strconv.ParseBool(config.Annotations["submariner.io/skip-cloud-prepare"]) {
```

tpantelis: Can't do exactly like that since `ParseBool` also returns an error if it's not "true" or "false" or allowed synonyms or if it's empty (ie annotation not present).  I think it's easiest to just accept "true" and anything else is false. This is just an undocumented backdoor.

skitt: Ah yes, it returns `(bool, error)`.

yboaron (approved):

PR #1672, opened 2024-10-08
Dockerfile: confirm microdnf update automatically

tpantelis:
microdnf gets stuck when the build runs interactively. Specifying "-y" avoids any prompting and allows the build to complete.


tpantelis:
@skitt This is what caused the konflux jobs to hang.

yboaron (approved):

sridhargaddam (approved):

skitt:
@tpantelis ah, same as #1603.

PR #1681, opened 2024-10-11
2.10: Bump submariner dependencies to 0.17.3

tpantelis:


skitt:
Weird, I would expect this to conflict now that #1687 has been merged...

tpantelis:
> Weird, I would expect this to conflict now that #1687 has been merged...

It looks like this PR is a subset so it's essentially a no-op.

PR #1732, opened 2024-11-12
Switch to a Go 1.22 builder for Konflux

skitt:


tpantelis (approved):

tpantelis:
We should be able to cheery pick to 2.12 and 2.10 branches

PR #1964, opened 2025-04-09
Bump golang-jwt/v5 to 5.2.2 to fix CVE-2025-30204

dfarrell07:
Change generated by:

    go get github.com/golang-jwt/jwt/v5@v5.2.2
    go mod tidy

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -85,7 +85,7 @@ require (
 	github.com/go-openapi/swag v0.23.0 // indirect
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
-	github.com/golang-jwt/jwt/v5 v5.2.1 // indirect
+	github.com/golang-jwt/jwt/v5 v5.2.2 // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/btree v1.1.3 // indirect
 	github.com/google/cel-go v0.22.0 // indirect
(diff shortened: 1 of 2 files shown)


dfarrell07:
We also have `github.com/golang-jwt/jwt v3.2.1+incompatible`:

```
[~/go/src/stolostron/submariner-addon]$ go list -mod=mod -m all|grep github.com/golang-jwt/jwt
github.com/golang-jwt/jwt v3.2.1+incompatible
```

It makes sense that it would be impacted by the CVE, but the CVE says `github.com/golang-jwt/jwt/v4 < 4.5.2` which is technically a different package. That pre-v4 package doesn't have a fix, so we would need to use a replace statement to try to force a newer version.

It comes to us from:

```
[~/go/src/stolostron/submariner-addon]$ go mod graph | grep -i github.com/golang-jwt/jwt 
github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute@v1.0.0 github.com/golang-jwt/jwt@v3.2.1+incompatible
github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork@v1.1.0 github.com/golang-jwt/jwt@v3.2.1+incompatible
```

skitt (changes requested):
v4 _doesn’t_ end up in the build (and neither does v3):

```
$ make
$ go version -m submariner | grep jwt
	dep	github.com/golang-jwt/jwt/v5	v5.2.1
```

It shows up in the `go list -m` because it’s part of the module tree and participates in MVS, but it gets pruned at the package level before the build. You can verify this by looking in `vendor` after `go mod vendor` too.

dfarrell07:
Dropped the v4 update, leaving the same v5 update commit.

PR #2046, opened 2025-05-07
Allow Dependabot to update operator-framework

dfarrell07:
We disabled this in #1634. There haven't been any updates to the module
on any branch since, although there have been framework releases and
Submariner releases.

Depends on #2045


skitt:
Missing these was a side-effect of moving the Operator SDK dependency out of the Submariner Operator’s main module… We’ll need to keep an eye on the interactions between the addon framework and the operator framework but that’s what CI is for.

/approve

PR #2103, opened 2025-05-21
Remove builder metadata

skitt:
We don't need to set labels or copy the license in the build container; only the image built in the last stage needs these pieces of information.


dfarrell07:
Cool, pleasantly surprised that works. I thought it was needed for both builds because the errors were duplicated twice with two different container IDs.

/lgtm

dfarrell07:
> Error allocating host: taskruns.tekton.dev "submariner-addon-acm-04217cf61dd0eaaa798fd5c922987b31-provision" already exists

Infra?

/retest

dfarrell07:
Same failure, although it's taking ages for it to actually fail and report back here.

PR #2114, opened 2025-05-21
Enable prefetch on the push job too

skitt:


dfarrell07:
Ah cool, I had sent https://github.com/stolostron/submariner-addon/pull/2115 as well.

/lgtm

PR #2131, opened 2025-06-04
Fix new Konflux EC violations

dfarrell07:


dfarrell07:
So far, EC job is saying it's canceled because the build job failed but the build job hasn't reported any error, it just hasn't started the build-image-index step.

dfarrell07:
The logs are already gone?? :thinking: 

/retest

dfarrell07:
> So far, EC job is saying it's canceled because the build job failed but the build job hasn't reported any error, it just hasn't started the build-image-index step.

The ~10m lag between the build-images step passing and the build-image-index step starting is a consistent bug.

aswinsuryan (approved):

PR #2133, opened 2025-06-06
Update Konflux task refs

dhaiducek:

PR #2134, opened 2025-06-07
Konflux test

dfarrell07:


dfarrell07:
This PR was to double-check EC after https://github.com/stolostron/submariner-addon/pull/2133. It's passing with no violations, just some warnings about warnings as quoted in https://github.com/stolostron/submariner-addon/pull/2133#issuecomment-2951590667.

PR #2168, opened 2025-06-13
Bump to Kubernetes 1.33 and Go 1.24

vthapar:
Depends on https://github.com/openshift/library-go/pull/1952


skitt:
This shouldn’t be necessary, we’ll get it as a side-effect of the library-go bump when that’s ready, or the Submariner bump following the K8s 1.33 merge there. (It’s best to bump K8s along with Submariner to ensure that all the dependencies are aligned with K8s deprecations and removals.)

vthapar:
> This shouldn’t be necessary, we’ll get it as a side-effect of the library-go bump when that’s ready, or the Submariner bump following the K8s 1.33 merge there. (It’s best to bump K8s along with Submariner to ensure that all the dependencies are aligned with K8s deprecations and removals.)

Will dependabot be enough or will we need some changes in our code to make the bump work?

skitt:
> > This shouldn’t be necessary, we’ll get it as a side-effect of the library-go bump when that’s ready, or the Submariner bump following the K8s 1.33 merge there. (It’s best to bump K8s along with Submariner to ensure that all the dependencies are aligned with K8s deprecations and removals.)
> 
> Will dependabot be enough or will we need some changes in our code to make the bump work?

Well, you had to change the Dockerfiles in this PR :wink:. See https://github.com/stolostron/submariner-addon/pull/2206 for the whole set of changes — you could have a Go 1.24 PR separately, then dependabot would push a Submariner PR but that wouldn’t deal with the library-go bump (since that’s digest-based).

PR #2293, opened 2025-07-09
Bump to Submariner 0.21.0-rc0

vthapar:
Also bump to Go 1.24


vthapar:
Closing this as main will get bump later when K8s 1.33 is also ready to bump. This Pr will be re-created directly on `release-2.14` with the fix for dependency causing failure.

PR #2365, opened 2025-08-26
Add support for configurable resource limits and requests

tpantelis:
...for the submariner add-on agent deployment following the guidelines described here:
https://github.com/stolostron/foundation-docs/blob/main/addon/addon-resource-requirements.md

This entails migrating to the `AgentAddonFactory` API which creates an `AgentAddon` implementation from various configuration options. This does all the work of creating the manifest resources from the template YAML files and handling custom `NodeSelector`, `Tolerations`, and `ResourceRequirements` configured in an AddOnDeploymentConfig. We just need to place the proper template substitution code in the deployment template YAML file.

The _manifests_ directory under _submarineraddonagent_ was split into two sub directories, _spoke_ and _hub_, to avoid the `AgentAddonFactory` code from including the hub-specific RBAC resources in the agent manifest.

Related to https://issues.redhat.com/browse/ACM-19017


pkg/hub/submarineraddonagent/agent.go
```diff
@@ -59,30 +43,19 @@ const (
...
-	"manifests/role.yaml",
-	"manifests/rolebinding.yaml",
-	"manifests/deployment.yaml",
-}
-
 var agentHubPermissionFiles = []string{
-	"manifests/hub_role.yaml",
-	"manifests/hub_rolebinding.yaml",
+	"manifests/hub/hub_role.yaml",
+	"manifests/hub/hub_rolebinding.yaml",
```
skitt, line 48: How about renaming these to avoid the `hub` stutter and match the `spoke` file names?

PR #2425, opened 2025-09-15
Switch to openshift-golang-builder in the regular build

skitt:


skitt:
This is just to see what breaks :wink:

PR #2456, opened 2025-09-30
Bump submariner to 0.20.2

yboaron:


yboaron:
cc @vthapar @tpantelis

PR #2682, opened 2026-04-28
Fix compile error in submariner addon

gparvin:
Fix the compile problem konflux identified with this component.

Refs:
 - https://redhat.atlassian.net/browse/ACM-32492


skitt:
/lgtm
/approve

Thanks!

PR #2683, opened 2026-04-28
Fix the konflux dockerfile to address golang version difference

gparvin:
This build was broken due to using go 1.24 in the Dockerfile, but the go module was setup for 1.25

Refs:
 - https://redhat.atlassian.net/browse/ACM-32496


skitt:
/lgtm
/approve

Thanks!

PR #2701, opened 2026-05-21
Use the CSRApprover method from addon framework

vthapar:
Addonframework has changed how it creates CSR. This causes a failure in csrApproveCheck().

Replace the custom ccsrApproveCheck with default provided by the framework. This will make sure that we don't need to keep the code in addon in sync with whatever changes framework makes in future.

Refer: https://github.com/open-cluster-management-io/addon-framework/commit/a60dd63a0abcbe82c8770a3dfba60d3cc8f86ed5

## Summary by CodeRabbit

* **Refactor**
  * Simplified CSR approval configuration and removed redundant CSR parsing and validation logic.
* **Tests**
  * Updated CSR-related tests and test data to include username handling and adjusted cases to match the new approval behavior.

### DIFF
--- a/pkg/hub/submarineraddonagent/agent.go
+++ b/pkg/hub/submarineraddonagent/agent.go
@@ -2,9 +2,7 @@ package submarineraddonagent
 
 import (
 	"context"
-	"crypto/x509"
 	"embed"
-	"encoding/pem"
 	goerrors "errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


skitt (approved):

vthapar:
@tpantelis csrTest in agent_test.go is failing. Considering we don't have it in our code anymore, won't be better to just remove the test? Else it will fail anytime addon framework changes CSR logic.

yboaron (approved):

PR #2709, opened 2026-06-01
Revert "Update the catalog to use new probe ports"

tpantelis:
Reverts stolostron/submariner-addon#2700

## Chores
- Updated health check probes to use HTTPS protocol on port 8443, replacing the previous HTTP configuration on port 8081, for both readiness and liveness checks.


vthapar:
Need revert of https://github.com/stolostron/submariner-addon/pull/2695 too. Both need to go in together or addon will fail to come up.

tpantelis:
Included with https://github.com/stolostron/submariner-addon/pull/2710

PR #2733, opened 2026-07-02
Remove legacy OLM catalog v0.3.0 and packagemanifests format

tpantelis:
This removes the old packagemanifests OLM catalog format (v0.3.0 from 2021) in favor of the modern bundle format. The repository has migrated to using only the bundle format with manifests/ directory.

The OLM catalog now uses only the bundle format with v0.4.0 in manifests/, which is consistent with the Makefile (CSV_VERSION=0.4.0) and metadata/annotations.yaml.


vthapar:
Changes look good, but will this need a backport? wondering if will impact backporting any changes in manifest files, shouldn't, just wondering.

vthapar (approved):

tpantelis:
> Changes look good, but will this need a backport? wondering if will impact backporting any changes in manifest files, shouldn't, just wondering

AFAICT, the v0.3.0 manifest file has been stale for 5 years or so and was possibly used, if at all, for migration from v0.3.0.

Issue #2726, opened 2026-06-23
Gateway labels not re-applied after node replacement during ROKS cluster upgrade


yboaron:
When cluster upgrades replace worker nodes (for example IBM ROKS 4.20→4.21) , Submariner gateway labels are lost and never restored,
breaking all cross-cluster connectivity.

Symptom:
 - Old gateway node deleted during upgrade
 - No nodes have submariner.io/gateway=true label
 - Connectivity broken until manual intervention

 Workaround:
triggers a manual reconciliation of submariner-addon controller at a time when the new nodes are ready resolve the issue, for example : 
```
  oc patch submarinerconfig -n <cluster-namespace> submariner \
    --type merge -p '{"spec": {"gatewayConfig": {"gateways": 1}}}'

```

PR #2737, opened 2026-07-27
Release 2.17 : fix gateway label restoration after node replacement

yboaron:
When cluster upgrades replace worker nodes, Submariner gateway labels were not being automatically restored on new nodes, breaking all cross-cluster connectivity until manual intervention.

This PR adds needsGatewayReconciliation() function that checks if the actual number of labeled gateway nodes matches the desired count. The sync is now performed when either:
1. The SubmarinerConfig spec has changed, OR
2. The actual gateway count doesn't match the desired count

This ensures automatic re-labeling when nodes are replaced during upgrade processes without requiring manual intervention.

Fixes: #2726

### DIFF
--- a/pkg/spoke/submarineragent/config_controller.go
+++ b/pkg/spoke/submarineragent/config_controller.go
@@ -231,9 +231,23 @@ func (c *submarinerConfigController) sync(ctx context.Context, syncCtx factory.S
 func (c *submarinerConfigController) syncConfig(ctx context.Context, recorder events.Recorder,
 	config *configv1alpha1.SubmarinerConfig,
 ) error {
-	if c.skipSyncingUnchangedConfig(config) {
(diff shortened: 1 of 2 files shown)


tpantelis:
@yboaron @vthapar **tide** says "_Not mergeable. Needs acknowledge-security-fixes-only label._". It seems the `release-2.17` branch is in a "security fixes only" mode, meaning it only accepts critical security patches. This PR is a bug fix (not a security fix), so tide requires explicit acknowledgment via the `acknowledge-security-fixes-only` label to bypass this restriction. Is this bug fix critical enough to do so?

yboaron:
> Is this bug fix critical enough to do so?

This bug broke Submariner connectivity after upgrading from ROKS 4.19 to 4.20, user says this is a common upgrade path, and the fix should be in an official release (they use 2.16), however, we do have WA, and it is also documented in KCS  https://access.redhat.com/solutions/7144422

PR #2746, opened 2026-08-19
Drop spoke-agent update/patch privileges for hub resources

tpantelis:
`SubmarinerConfig.spec.imagePullSpecs` (7 free-form image refs) and `spec.subscriptionConfig` flow with no allow-list, registry pin, or digest check via `brokerinfo.applySubmarinerImageConfig` into `imageOverrides` on the spoke's `Submariner` CR (delivered via `ManifestWork`). **submariner-gateway** / **route-agent** run hostNetwork + NET_ADMIN + privileged. The hub-side agent Role gave each spoke's addon-agent `ServiceAccount` update/patch on submarinerconfigs in its own per-cluster namespace, so a compromised spoke could rewrite its own `SubmarinerConfig.spec `on the hub and have the hub deploy an attacker-controlled image as a privileged host-network pod on the spoke's gateway nodes.

The spoke agent only reads `SubmarinerConfig` and writes the `/status` subresource. Drop update/patch on the main resource; keep get/list/watch and the existing /status and /finalizers rules.

Same with `ManagedClusterAddOn`.

## Summary by CodeRabbit

* **Security Improvements**
  * Restricted the hub agent’s access to configuration resources to read-only operations.
  * Preserved permission to update resource status and finalizers.
  * Added documentation clarifying protected configuration fields.

### DIFF
--- a/pkg/hub/submarineraddonagent/manifests/hub/role.yaml
+++ b/pkg/hub/submarineraddonagent/manifests/hub/role.yaml
@@ -4,20 +4,26 @@ metadata:
   name: open-cluster-management:submariner-addon:agent
   namespace: {{ .ClusterName }}
 rules:
-# Allow submariner-addon agent to operator managedclusteraddons on the hub cluster
+# Allow submariner-addon agent to read managedclusteraddons on the hub cluster.
+# The agent only writes status (UpdateStatus in pkg/addon/update_status.go).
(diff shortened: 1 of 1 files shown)


vthapar (approved):
Not  a blocking comment, but spoke agent also has finalizer update permissions that it doesn't need. Can address it in a follow up PR.

